PR #66, opened 2025-06-30
Feat/visibility

julius-malcovsky:
During validation of a Rover resource in the webhook, make a call to the API domains webhook to validate the ApiSubscription. This will prevent creating stuck resources that would always be rejected by the target domains webhook.

### DIFF
--- a/admin/api/v1/zone_types.go
+++ b/admin/api/v1/zone_types.go
@@ -10,6 +10,13 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+type ZoneVisibility string
+
+const (
+	ZoneVisibilityWorld      ZoneVisibility = "World"
+	ZoneVisibilityEnterprise ZoneVisibility = "Enterprise"
+)
+
 type RedisConfig struct {
 	Host     string `json:"host"`
 	Port     int    `json:"port"`
@@ -54,6 +61,7 @@ type ZoneSpec struct {
 	Gateway          GatewayConfig          `json:"gateway"`
 	Redis            RedisConfig            `json:"redis"`
 	TeamApis         *TeamApiConfig         `json:"teamApis,omitempty"`
+	Visibility       ZoneVisibility         `json:"visibility"`
 }
 
 type Links struct {
--- a/api/internal/webhook/v1/apisubscription_webhook.go
+++ b/api/internal/webhook/v1/apisubscription_webhook.go
@@ -1,0 +1,180 @@
+// SPDX-FileCopyrightText: 2025 Deutsche Telekom IT GmbH
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package v1
+
+import (
+	"context"
+	"fmt"
+	adminv1 "github.com/telekom/controlplane/admin/api/v1"
+	cclient "github.com/telekom/controlplane/common/pkg/client"
+	"github.com/telekom/controlplane/common/pkg/config"
+	"github.com/telekom/controlplane/common/pkg/util/labelutil"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	apiv1 "github.com/telekom/controlplane/api/api/v1"
+)
+
+// nolint:unused
+// log is for logging in this package.
+var apisubscriptionlog = logf.Log.WithName("apisubscription-resource")
+
+// SetupApiSubscriptionWebhookWithManager registers the webhook for ApiSubscription in the manager.
+func SetupApiSubscriptionWebhookWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewWebhookManagedBy(mgr).For(&apiv1.ApiSubscription{}).
+		WithValidator(&ApiSubscriptionCustomValidator{mgr.GetClient()}).
+		WithDefaulter(&ApiSubscriptionCustomDefaulter{}).
+		Complete()
+}
+
+// +kubebuilder:webhook:path=/mutate-api-cp-ei-telekom-de-v1-apisubscription,mutating=true,failurePolicy=fail,sideEffects=None,groups=api.cp.ei.telekom.de,resources=apisubscriptions,verbs=create;update,versions=v1,name=mapisubscription-v1.kb.io,admissionReviewVersions=v1
+
+// ApiSubscriptionCustomDefaulter struct is responsible for setting default values on the custom resource of the
+// Kind ApiSubscription when those are created or updated.
+//
+// NOTE: The +kubebuilder:object:generate=false marker prevents controller-gen from generating DeepCopy methods,
+// as it is used only for temporary operations and does not need to be deeply copied.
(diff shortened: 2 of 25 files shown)


admin/api/v1/zone_types.go
```diff
@@ -54,6 +61,7 @@ type ZoneSpec struct {
 	Gateway          GatewayConfig          `json:"gateway"`
 	Redis            RedisConfig            `json:"redis"`
 	TeamApis         *TeamApiConfig         `json:"teamApis,omitempty"`
+	Visibility       ZoneVisibility         `json:"visibility"`
```
ron96G, line 64: Maybe add a small comment what this attribute entails

api/internal/webhook/v1/apisubscription_webhook.go
```diff
@@ -0,0 +1,175 @@
...
+		return nil, fmt.Errorf("expected a ApiSubscription object but got %T", obj)
+	}
+	apisubscriptionlog.Info("Validation for ApiSubscription upon deletion", "name", apisubscription.GetName())
+
+	// todo handle deletion
+
+	return nil, nil
+}
+
+func getEnvironment(object client.Object) (string, bool) {
```
ron96G, line 112: Already defined in "github.com/telekom/controlplane/common/pkg/controller"

api/internal/webhook/v1/apisubscription_webhook.go
```diff
@@ -0,0 +1,175 @@
...
+	labels := object.GetLabels()
+	if labels == nil {
+		return "", false
+	}
+	e, ok := labels[config.EnvironmentLabelKey]
+	return e, ok
+}
+func validateCreateOrUpdate(ctx context.Context, c client.Client, sub apiv1.ApiSubscription) (admission.Warnings, error) {
```
ron96G, line 120: apiv1.ApiSubscription should be a pointer

api/internal/webhook/v1/apisubscription_webhook.go
```diff
@@ -0,0 +1,175 @@
...
+	err := scopedClient.List(ctx, apiExposureList,
+		client.MatchingLabels{apiv1.BasePathLabelKey: labelutil.NormalizeValue(sub.Spec.ApiBasePath)},
+		client.MatchingFields{"status.active": "true"})
+
+	if err != nil {
+		apisubscriptionlog.Error(err, "unable to list ApiExposure", "name", sub.Spec.ApiBasePath)
+		return nil, apierrors.NewNotFound(schema.GroupResource{Group: apiv1.GroupVersion.Group, Resource: "ApiExposure"}, "Active Api Exposure for this subscription not found due to error")
+	}
+
+	if len(apiExposureList.Items) == 0 {
```
ron96G, line 140: len(apiExposureList.Items) == 0 is not an error case. It is accepted that an an ApiSub can exist without an ApiExp. It will be set to blocked and processed when the ApiExp is created

api/internal/webhook/v1/webhook_suite_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+// getFirstFoundEnvTestBinaryDir locates the first binary in the specified path.
+// ENVTEST-based tests depend on specific binaries, usually located in paths set by
+// controller-runtime. When running tests directly (e.g., via an IDE) without using
+// Makefile targets, the 'BinaryAssetsDirectory' must be explicitly configured.
+//
+// This function streamlines the process by finding the required binaries, similar to
+// setting the 'KUBEBUILDER_ASSETS' environment variable. To ensure the binaries are
+// properly set up, run 'make setup-envtest' beforehand.
+func getFirstFoundEnvTestBinaryDir() string {
```
ron96G, line 139: Was this generated or did you add this?

common/pkg/client/mocks/client.go
ron96G, line 1: Why is sigs.k8s.io/controller-runtime/pkg/client mocked? Should ScopedClient/JanitorClient be enough? Also how was this generated? Where is the annotation?

common/pkg/client/mocks/janitor_client.go
ron96G, line 1: same as mocks/client.go

rover/internal/handler/handler_suite_test.go
```diff
@@ -0,0 +1,130 @@
...
+		// the desired error
+		statusErr := apierrors.StatusError{}
+		statusErr.ErrStatus.Reason = metav1.StatusReasonBadRequest
+		// Expect the client Create call to succeed
+		mockClient.EXPECT().
+			CreateOrUpdate(ctx, gomock.AssignableToTypeOf(apiSub), gomock.Any()).
+			Return(controllerutil.OperationResultNone, &statusErr) // the created doesnt matter here
+
+		err := apihandler.HandleSubscription(ctx, mockClient, rover, roverSub)
+		AssertConditionBlockedPresence(rover.GetConditions(), "", false)
```
ron96G, line 106: Why is the expectedMessage empty?

rover/internal/handler/rover/api/subscription.go
```diff
@@ -63,12 +65,25 @@ func HandleSubscription(ctx context.Context, c client.JanitorClient, owner *rove
...
-
 		return nil
 	}
 
 	_, err := c.CreateOrUpdate(ctx, apiSubscription, mutator)
-	if err != nil {
+
+	// many different errors can occur, so lets handle them
+	var statusErr *apierrors.StatusError
+	if errors.As(err, &statusErr) {
```
ron96G, line 75: Does it make sense to move this to a central error-handling in common, e.g. in the ScopedClient. Map these k8s error to internal errors?

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+<!--
```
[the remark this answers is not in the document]
julius-malcovsky: removed, but would be nice if this could be configured somewhere, when i create new files, reuse will always update the changelog as well ...

api/internal/webhook/v1/apisubscription_webhook.go
julius-malcovsky: my mistake - fixed

api/internal/webhook/v1/apisubscription_webhook.go
julius-malcovsky: ATM only rover is creating new ApiSubscriptions, if this happens, rover handles this webhook response and sets the rover to blocked (via condition) - is that ok or do you want to rework it (if so lets have a chat)

api/internal/webhook/v1/webhook_suite_test.go
julius-malcovsky: generated

admin/api/v1/zone_types.go
julius-malcovsky: updated

rover/internal/handler/handler_suite_test.go
julius-malcovsky: the last parameter is false, so im not expecting the condition to be present, thus the message made no sense to me .. did I missunderstand the usage of this ?

api/internal/webhook/v1/apisubscription_webhook.go
julius-malcovsky: I believe there might be a more complex problem connected to this. 

What you wrote makes sense, I have no problem with that and I agree. Then the webhook would do, at the moment, just the visibility rules check. For that we need to have the active exposure to get its visibility of it so we can compare the subscription values. That creates the question, what should the webhook reply if it cannot get some data thats required for it to make a decision. 

Perhaps we can talk about another mechanism how to do these things. Webhooks are very good when all data is present, but in this case they feel heavy to use. Lets talk :)

julius-malcovsky:
was replaced by a different implementation, closing the PR, keeping the branch (might be useful for webhook development)

PR #81, opened 2025-07-10
feat(secret-manager): add cache metrics for cache hits and misses

stefan-ctrl:


secret-manager/cmd/server/server.go
```diff
@@ -121,6 +130,10 @@ func main() {
 
 	ctx := cs.SignalHandler(context.Background())
 
+	// Register metrics
+	cache.RegisterMetrics(prometheus.DefaultRegisterer)
+	bouncer.RegisterMetrics(prometheus.DefaultRegisterer)
```
BjoernKarma, line 135: Does it make sense to move this to the `metrics.go` file where the metrics are also defined? Then all the logic concerning metrics would be in the same place. In `identity` we do this inside an `init` function:
```
func init() {
	// Register custom metrics with the global prometheus registry
	metrics.Registry.MustRegister(keycloakRequests, keycloakRequestsFailures, keycloakRequestDuration)
}
```

secret-manager/pkg/backend/cache/backend.go
```diff
@@ -40,8 +40,15 @@ func (c *CachedBackend[T, S]) ParseSecretId(raw string) (T, error) {
...
-		return item.Value(), nil
+	if item, ok := c.Cache.Get(id.String()); ok {
+		if !item.Expired() {
+			RecordCacheHit()
+			return item.Value(), nil
+		} else {
+			RecordCacheMiss("expired")
+		}
+	} else {
+		RecordCacheMiss("not_found")
```
BjoernKarma, line 51: Is this new `else` block required or can the `RecordCacheMiss("not_found")` just be added to the code block below that handles the cache miss?

secret-manager/pkg/backend/cache/metrics.go
```diff
@@ -0,0 +1,40 @@
...
+		prometheus.CounterOpts{
+			Name: "cache_access_total",
+			Help: "Total number of cache access attempts",
+		},
+		[]string{"result", "reason"},
+	)
+)
+
+// RegisterMetrics registers all cache-related metrics with Prometheus
+func RegisterMetrics(reg prometheus.Registerer) {
```
BjoernKarma, line 26: Do we need this sync.Once for registering metrics? I am just wondering, because at least in identity we are not using it. If we need it, then we might need to add it in other places as well

secret-manager/pkg/backend/cache/metrics_test.go
```diff
@@ -0,0 +1,120 @@
+// Copyright 2025 Deutsche Telekom IT GmbH
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package cache_test
```
BjoernKarma, line 5: I like that you have added unit tests for the metrics 👍Just wondering if we are testing our configuration with this test or the prometheus framework itself? Maybe an addition/alternative would be to verify in the `backend_test.go` that the `RecordCache...` functions are called or how often they are called for the appropriate test cases.
What do you think?

BjoernKarma (commented):
Looks good to me. I have just some minor questions/remarks :-)

secret-manager/pkg/backend/cache/metrics.go
stefan-ctrl: It's not required, but good practice since otherwise, the Prometheus library will panic if MustRegister is called multiple times.

I think `sync.once` it is used in identity under the hood: 

https://github.com/telekom/controlplane/blob/c71f1fe04cc976e0c9da578737c78145d524fa0e/identity/cmd/main.go#L72

https://github.com/telekom/controlplane/blob/c71f1fe04cc976e0c9da578737c78145d524fa0e/secret-manager/api/metrics/metrics.go#L9

https://github.com/telekom/controlplane/blob/c71f1fe04cc976e0c9da578737c78145d524fa0e/common-server/pkg/client/metrics.go#L37-L41

secret-manager/pkg/backend/cache/metrics_test.go
stefan-ctrl: True. It's a somewhat complicated setup to verify that the methods inside `metrics.go` work. However, I don't think, we test the Prometheus framework itself, but rather the units of `metrics.go`, that the right counters are increased as expected.

I like the approach to test them in `backend_test.go` as well.  Will add them as an addition.

secret-manager/pkg/backend/cache/metrics.go
BjoernKarma: Okay, I had the hope that the register function took care of this. But then it is a good idea to make sure it is called only once. At least in the keycloak metrics in identity we might need it as well then (https://github.com/telekom/controlplane/blob/c71f1fe04cc976e0c9da578737c78145d524fa0e/identity/pkg/keycloak/custom_metrics.go).

secret-manager/pkg/backend/cache/backend.go
BjoernKarma: Yes, you are right. It is needed to distinguish the expired case to not count the cache miss twice. If it is not needed right now, I think it would be okay to leave out the distinction for now. We can still add it later if we need it. It depends on the goal of the investigation. In the ticket it looks like for the time being, we want to distinguish between cache hit and cache miss.

secret-manager/pkg/backend/cache/backend.go
```diff
@@ -40,8 +41,15 @@ func (c *CachedBackend[T, S]) ParseSecretId(raw string) (T, error) {
 
 func (c *CachedBackend[T, S]) Get(ctx context.Context, id T) (res S, err error) {
 	log := logr.FromContextOrDiscard(ctx)
-	if item, ok := c.Cache.Get(id.String()); ok && !item.Expired() {
-		return item.Value(), nil
+	if item, ok := c.Cache.Get(id.String()); ok {
+		if !item.Expired() {
```
ron96G, line 45: Can this ever happen? Looking at the code in [here ](https://github.com/telekom/controlplane/blob/5ddda4f4283d0ad0078ecec78bcc4a78c207b37c/secret-manager/pkg/backend/cache/cache.go#L39). It already checks if its expired and then returns nil. You could `return item, false` but your check `if ok` already excludes that

stefan-ctrl: Nope, can never happen, and have discussed this here as well: https://github.com/telekom/controlplane/pull/81#discussion_r2204424796

I'll remove this metrics call for now.

stefan-ctrl: refactored in https://github.com/telekom/controlplane/pull/81/commits/e76309f6179051e900f32d8c3c347b35244be561

stefan-ctrl:
Removed the interface. Should now be ready to be merged

ron96G (approved):

PR #42, opened 2025-05-28
chore(deps): upgrade to vektra/mockery/v3

stefan-ctrl:
Migration steps:

https://vektra.github.io/mockery/v3.2/v3/#mockery-migrate:

```console
mockery migrate --config tools/mockery.yaml
```

https://vektra.github.io/mockery/v3.2/v3/#with-expecter
> ```with-expecter```
> In v3, this parameter has been removed. testify-style mocks will always generate expecter methods.

---

This PR also hopes to solve the issue with the flaky tests coming from our GingGo x Mockery.  
Example: #31: https://github.com/telekom/controlplane/actions/runs/15293540857/job/43017369817#step:7:26

PR #86, opened 2025-07-14
test(identity): fix flaky tests for keycloak-client-mock cleanup

ron96G:


stefan-ctrl (commented):
LGMT 

Have you tested the Pipeline/Jobs by running it multiple times to verify that it works? If so, I would approve it.

Tried to solve this by upgrade mockery, but seems like I was on the "Holzpfad" (see #42)

PR #88, opened 2025-07-15
feat: added basic-auth feature

ron96g:


api/config/crd/bases/api.cp.ei.telekom.de_remoteapisubscriptions.yaml
ellevaoche, line 1: Do we want basiCAuth Support in RemoteApiSubscriptions?

ellevaoche: Okay, then it's fine :-)

PR #95, opened 2025-07-21
feat: Added Private Key JWT feature

iagarba:


gateway/internal/features/feature/external_idp.go
```diff
@@ -123,13 +123,25 @@ func extendOauth(ctx context.Context, in plugin.OauthCredentials, providerSettin
...
+
+	if clientKey != "" {
+		clientKey, err = secretManagerApi.Get(ctx, clientKey)
 		if err != nil {
 			return in, err
 		}
+
+		in.ClientKey = clientKey
+	} else {
+		if secret != "" {
```
stefan-ctrl, line 136: This could be combined in an else if statement

rover/api/v1/rover_types.go
```diff
@@ -545,4 +545,7 @@ type OAuth2ClientCredentials struct {
 	// ClientSecret is the secret associated with the client ID
 	// +kubebuilder:validation:Optional
 	ClientSecret string `json:"clientSecret"`
+	// ClientKey is the private key associated with the client ID
+	// +kubebuilder:validation:Optional
+	ClientKey string `json:"clientKey"`
```
stefan-ctrl, line 550: Perhaps we should add the check that `key` or `secret` could be used for rover validation as well, as you did in `api` domain

iagarba: Done!

PR #110, opened 2025-07-29
feat(file-manager): Add deployment config for file-manager

BjoernKarma:
Prepare deployment config to enable cluster deployment:
- add certmanager for deployment
- add filemgr-token for AWS STS for deployment


file-manager/config/default/deployment_patch.yaml
```diff
@@ -0,0 +1,34 @@
...
+      sources:
+        - serviceAccountToken:
+            path: filemgr-token
+            expirationSeconds: 600
+            audience: sts.amazonaws.com
+- op: add
+  path: /spec/template/spec/containers/0/volumeMounts/-
+  value:
+    name: filemgr-token
+    mountPath: /var/run/files/filemgr
```
ron96G, line 33: copy/paste mistake: "/var/run/files/filemgr" should be "/var/run/secrets/filemgr"...

Also, maybe filemgr -> file-manager.... Not sure if the abbreviation is super nice

BjoernKarma: Adjusted names and paths for token

PR #114, opened 2025-08-01
feat/rate-limiting

stefan-ctrl:


ron96g:
@stefan-ctrl the rate-limiting options are needed for consumer, consumerDefault  and provider. If none for consumer are set, then the provider act as default. If provider is not set, cluster-defaults as used.

PR #136, opened 2025-08-19
Chore/bump go 1.24.6 with chore

stefan-ctrl:


.github/workflows/reusable-go-ci.yaml
```diff
@@ -232,6 +232,7 @@ jobs:
       - name: Run govulncheck
         uses: golang/govulncheck-action@b625fbe08f3bccbe446d94fbf87fcc875a4f50ee # pin@1.0.4
         with:
+          go-version-file: "${{ inputs.module }}/go.mod"
```
ron96g, line 235: Warning: Both go-version and go-version-file inputs are specified, only go-version will be used

ron96g: see https://github.com/golang/govulncheck-action/blob/master/action.yml ... This action sets a default "stable" for the go-version, so this value is actually never used?

ron96g (commented):
Looks good. Except the 1 new warning in the pipeline logs

.github/workflows/reusable-go-ci.yaml
stefan-ctrl: We could try to safely remove it (so just leave it as it is before this commit). 

However, I would prefer to pin the version to the module version, since now "stable" would refer to `1.25.0`, which could lead to errors in the future.

stefan-ctrl: Removed in 6c1695f0509445e02d744d913804b1cc7f1d0e1f

PR #166, opened 2025-09-16
ci(codeql): Aggregate codeql reports from all modules

BjoernKarma:


BjoernKarma:
Aggregation is not needed. Fix will be applied with a different approach. PR can be closed

PR #178, opened 2025-09-24
fix: resolve reconciler loops in organization + identity

stefan-ctrl:
# Fix

## Identity Domain Fix
- Removed unnecessary status processing updates: Eliminated the SetStatusProcessing(client) call at the beginning of the 
CreateOrUpdate function in the client handler, preventing unnecessary status updates that were triggering reconciliation loops.

## Organization Domain Fix
- Commented out processing condition: Disabled the teamObj.SetCondition(condition.NewProcessingCondition("Ready", "Team is being processed")) call in the team handler, preventing status updates that were contributing to the reconciliation loop between organization and identity domains.

## Metrics
- add missing common-server metrics-client integration: Example: `http_client_request_duration_seconds_count{client="identity"}`

# Refactor
- Setting status in identity requires fewer parameters


ron96g (changes requested):
LGTM just the minor change with the oldSecretRef. I think identity could use a refactoring in general but thats out of scope for this task

PR #189, opened 2025-10-02
fix(install): correctly add labels to label-selector in servicemonitor

ron96g:


rover-server/config/default/kustomization.yaml
```diff
@@ -7,9 +7,14 @@ namespace: controlplane-system
 
 # Labels to add to all resources and selectors.
 labels:
-  - includeSelectors: true
-    pairs:
-      domain: rover
+- includeSelectors: true
+  pairs:
+    domain: rover
```
BjoernKarma, line 12: Is it okay that rover and rover-server use the same domain?

BjoernKarma (approved):
LGTM
Just one small question

rover-server/config/default/kustomization.yaml
BjoernKarma: 👍

PR #254, opened 2026-02-03
feat(tool): improved e2e-tester with improved validation and json-schema support; external suite configuration; better obfuscator patterns; more convenient run-policy

ron96g:


tools/e2e-tester/pkg/config/config.go
```diff
@@ -6,39 +6,94 @@ package config
...
+const (
+	// RunPolicyNormal - test runs when prior tests passed, skipped when prior failed.
+	RunPolicyNormal RunPolicy = "normal"
+
+	// RunPolicyCritical - test runs when prior passed, aborts suite on ERROR status.
+	RunPolicyCritical RunPolicy = "critical"
+
+	// RunPolicyAlways - test always runs regardless of prior failures (for cleanup).
+	RunPolicyAlways RunPolicy = "always"
+)
```
stefan-ctrl, line 29: Would prefer more speakable names for the config, that directly communicate its purpose


```go
const (
	// RunPolicySkippable - test runs when prior tests passed, skipped when prior failed.
	RunPolicyRunOnSuccess RunPolicy = "RunOnSuccess " //could be default

	// RunPolicyFailFast - test runs when prior passed, aborts suite on ERROR status.
	RunPolicyFailFast RunPolicy = "FailFast"

	// RunPolicyAlways - test always runs regardless of prior failures (for cleanup).
	RunPolicyAlways RunPolicy = "always"
)
```

tools/e2e-tester/pkg/environment/manager.go
```diff
@@ -14,10 +14,15 @@ import (
 	"go.uber.org/zap"
 )
 
+const (
+	envPrefix = "env://"
+)
```
stefan-ctrl, line 19: Why is `//` needed as an additional prefix for the environment? This is a breaking change.

tools/e2e-tester/pkg/report/reporter.go
```diff
@@ -35,7 +35,7 @@ type TestCaseResult struct {
 	Error          error
 	ComparisonDiff string
 	Environment    string
-	MustPass       bool
+	RunPolicy      string // Execution policy: "normal", "critical", "always"
```
stefan-ctrl: Should be of type `type RunPolicy string` to not repeat in comments which values are allowed

stefan-ctrl (changes requested):

stefan-ctrl (dismissed):

tools/e2e-tester/pkg/command/executor.go
```diff
@@ -54,7 +58,11 @@ func (e *RoverCtlExecutor) Execute(ctx context.Context, cmdStr string, params ma
 
 	// Set up environment variables
 	if e.environment.Token != "" {
-		cmd.Env = append(cmd.Env, fmt.Sprintf("ROVER_TOKEN=%s", e.environment.Token))
+		cmd.Env = append(cmd.Env, fmt.Sprintf("%s=%s", roverTokenEnvKey, e.environment.Token))
```
julius-malcovsky: why is the rover token treated separately? wouldnt the loop below set it ? 🤔

tools/e2e-tester/examples/basic-config.yaml
```diff
@@ -76,38 +47,38 @@ suites:
...
         compare: true
 
       - name: "apply-config"
         environment: team-b
         description: |
-          Applying this will result in a blocked API as the same API path 
+          Applying this will result in a blocked API as the same API path
           is already exposed by team-a
-        must_pass: false
+        # run_policy: RunOnSuccess (default)
```
julius-malcovsky, line 65: cosmetic, but sometimes even default values make sense. personal opinion but they can reduce the mental load. "what i am looking at right now is the configuration, no need to keep in mind every other value not present because its either here or it doesnt exist" ... just a thought :) feel free to resolve

tools/e2e-tester/pkg/config/config.go
```diff
@@ -6,40 +6,110 @@ package config
...
+		if strings.EqualFold(string(p), string(valid)) {
+			return true
+		}
+	}
+	return false
+}
+
+// Normalize returns the canonical PascalCase form of the policy.
+// Returns the input unchanged if it doesn't match any valid policy.
+func (p RunPolicy) Normalize() RunPolicy {
```
julius-malcovsky, line 46: it this maybe too forgiving ? maybe i misunderstood, but it feels like - we have enum but its ok if you dont care about them - we will fix the value for you

tools/e2e-tester/pkg/config/config.go
```diff
@@ -65,48 +135,89 @@ func (s Suite) GetName() string {
...
+	return ValidateConfig(c)
+}
 
-	// Ensure at least one case per suite and at least one environment per suite
-	for _, suite := range c.Suites {
-		if len(suite.Cases) == 0 {
-			return fmt.Errorf("suite %s must have at least one case", suite.Name)
+func (c *Config) LoadSuites(configDir string) error {
+	for i := range c.Suites {
+		if c.Suites[i].IsExternal() {
```
julius-malcovsky, line 215: will this only load external suits ? if thats the intent I would suggest renaming the func to reflect the purpose

tools/e2e-tester/pkg/config/validator.go
```diff
@@ -0,0 +1,218 @@
...
+		return // Shouldn't happen, but be defensive
+	}
+	if cfg.URL == "" && cfg.Binary == "" {
+		sl.ReportError(cfg.URL, "Snapshotter", "snapshotter", "url_or_binary_required", "")
+	}
+}
+
+// validateConfig performs all Config-level validations including
+// cross-reference checks and uniqueness constraints.
+func validateConfig(sl validator.StructLevel) {
```
julius-malcovsky: validateConfig + ValidateConfig can be a bit confusing - can we maybe rename them ?

tools/e2e-tester/pkg/config/validator.go
```diff
@@ -0,0 +1,218 @@
...
+		return "Snapshotter: either 'url' or 'binary' must be specified"
+	case "url":
+		return fmt.Sprintf("%s must be a valid URL (got: '%v')", field, e.Value())
+	case "env_reference":
+		return fmt.Sprintf("%s references unknown environment '%s'", field, e.Param())
+	case "unique_env_name":
+		return fmt.Sprintf("Environments must have unique names (duplicate: '%s')", e.Param())
+	case "unique_suite_name":
+		return fmt.Sprintf("Suites must have unique names (duplicate: '%s')", e.Param())
+	case "gte":
```
julius-malcovsky, line 211: didnt check, just asking - we only have gte? no ge or lte etc? is this the only "short" we use?

tools/e2e-tester/pkg/obfuscator/patterns.go
```diff
@@ -52,6 +52,19 @@ var (
...
+		{
+			Pattern: `trd_.+`,
+			Replace: "trd_OBFUSCATED",
+		},
+
+		{
+			Pattern: "uid.*",
+			Replace: "uid_OBFUSCATED",
+		},
+		{
```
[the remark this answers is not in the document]
julius-malcovsky: valid point - please consider it

tools/e2e-tester/pkg/runner/runner.go
```diff
@@ -323,16 +329,17 @@ func (r *Runner) Run(ctx context.Context) (*report.FinalReport, error) {
 
 		// Check if we should continue on failure
 		if !r.continueOnFail {
-			// Check if any must-pass test cases failed with error status
+			// Check if any critical-policy test cases had ERROR status
 			hasCriticalFailure := false
 			for _, tc := range result.Cases {
-				if tc.MustPass && tc.Status == report.StatusError {
+				if tc.RunPolicy == config.RunPolicyFailFast && tc.Status == report.StatusError {
```
julius-malcovsky: in config there were convenienvće funcs for run policies. would suggest being consistent and usin it here too

julius-malcovsky (dismissed):
mostly cosmetic, otherwise lgtm

tools/e2e-tester/pkg/command/executor.go
stefan-ctrl: Not really, since TOKEN and Variables are comming from different configurations sections. Rover Token is mandatory, wherease the other env vars are optional (hence it is a list).

stefan-ctrl: Not really, since TOKEN is a required field and hence not part fo the loop below. 

THe loop below is intended vor optional env vars, that may be required by other tools (such as snappshotter)

tools/e2e-tester/pkg/config/config.go
stefan-ctrl: True, but here it is intended to limit the possible values while not enforcing certain cases such as lowercase only, UPPERCASE. 

Since this is not customer-facing and for our internal testing and development, I'd prefer to be flexible here.

tools/e2e-tester/pkg/config/validator.go
stefan-ctrl: Yes, right now it is the only short we use,  is basically the same as lte just the other way arround. I'll leave it as is for now. BUt we ca easily extend it if the need raises :)

julius-malcovsky (approved):
Lovely

PR #286, opened 2026-03-18
Feature/dtc

julius-malcovsky:


gateway/internal/handler/realm/routes.go
```diff
@@ -61,14 +61,24 @@ func CreateRoute(ctx context.Context, realm *gatewayv1.Realm, routeType RouteTyp
...
+			Port:      gatewayv1.GetPortOrDefaultFromScheme(parsedUrl),
+			Path:      path.Join(parsedUrl.Path, fmt.Sprintf(cfg.DownstreamPathFormat, realm.Name)),
+			IssuerUrl: "",
+		})
 	}
 
 	mutator := func() error {
-		err := controllerutil.SetControllerReference(realm, route, c.Scheme())
-		if err != nil {
+		if err := controllerutil.SetControllerReference(realm, route, c.Scheme()); err != nil {
```
lukas016: optional: i don't recommend use the same variable name in nested function because it can be very easy be included over closures and it complicated review because difference is almost just one character := vs =. I prefer use custom names as locErr (local Error) or something else.

ron96g (changes requested):
Looks good so far. I already have plans in mind to overengineer this to the moon so maybe you have to stop me but please verify the comments.

PR #302, opened 2026-03-30
Feat/roadmap

julius-malcovsky:
Please see /docs for details - after the review we can remove these files.


ron96g (changes requested):
Will review further. For now, some open questions.

docs/roadmap-feature.md
```diff
@@ -0,0 +1,450 @@
+# Roadmap CRD Feature
```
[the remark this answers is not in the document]
julius-malcovsky: thanks for reminding me - removed

rover-server/api/openapi.yaml
```diff
@@ -43,6 +43,8 @@ tags:
     description: Manage EventSpecifications
   - name: Rover
     description: Manage Rovers
+  - name: Roadmap
+    description: Manage Roadmaps
```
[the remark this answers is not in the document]
julius-malcovsky: I would keep the new ones, since they are more flexible. Perhaps we could add special endpoints for backwards compatibility. they would accept the old format and map it into the new one. and we can announce the deprecation of the old ones, have the new ones and support the migration :)

rover-server/internal/controller/roadmap.go
```diff
@@ -0,0 +1,367 @@
...
+	roadmap.Spec.Hash = fileAPIResp.FileHash
+	EnsureLabelsOrDie(ctx, roadmap)
+
+	err = r.Store.CreateOrReplace(ctx, roadmap)
+	if err != nil {
+		return res, err
+	}
+
+	return r.Get(ctx, resourceId)
+}
```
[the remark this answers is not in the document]
julius-malcovsky: adjusted

rover-server/internal/controller/roadmap.go
```diff
@@ -0,0 +1,367 @@
...
+
+		var items []api.RoadmapItem
+		err = json.Unmarshal(b, &items)
+		if err != nil {
+			return nil, problems.InternalServerError("Failed to unmarshal roadmap items", err.Error())
+		}
+
+		// Create resource ID
+		resourceId := roadmap.Namespace[len(security.PrefixFromContext(ctx)):] + "--" + roadmap.Name
+		if len(roadmap.Namespace) > 0 && roadmap.Namespace[0:len(security.PrefixFromContext(ctx))] != security.PrefixFromContext(ctx) {
```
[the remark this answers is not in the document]
julius-malcovsky: ooooops :) 

resourceId is now handled properly - like ApiSpecification / EventSpecification

rover/internal/webhook/v1/roadmap_webhook.go
```diff
@@ -0,0 +1,83 @@
...
+func (v *RoadmapCustomValidator) ValidateUpdate(ctx context.Context, _ *roverv1.Roadmap, roadmap *roverv1.Roadmap) (admission.Warnings, error) {
+	return v.ValidateCreateOrUpdate(ctx, roadmap)
+}
+
+// ValidateDelete implements webhook.CustomValidator so a webhook will be registered for the type Roadmap.
+func (v *RoadmapCustomValidator) ValidateDelete(ctx context.Context, roadmap *roverv1.Roadmap) (admission.Warnings, error) {
+	return nil, nil
+}
+
+func (v *RoadmapCustomValidator) ValidateCreateOrUpdate(ctx context.Context, roadmap *roverv1.Roadmap) (admission.Warnings, error) {
```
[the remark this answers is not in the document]
julius-malcovsky: I think a loose coupling is better here. It give you the possibility to provider a "future" roadmap for something youre just working on. Also, it would create an order dependency, or would add complexity with blocked statuses and watchers etc. I liked the idea of Roadmaps being lightweight and be used as non functional metadata. Plus they are optional. 

What we could do is to return something to the user if they are referring to something that doesnt exist. Perhaps its intended and maybe its a typo and they would correct it.

rover/api/v1/roadmap_types.go
```diff
@@ -0,0 +1,120 @@
...
+
+	// Description provides detailed information about the roadmap item
+	Description string `json:"description"`
+
+	// TitleUrl is an optional URL link for the roadmap item title
+	TitleUrl string `json:"titleUrl,omitempty"`
+}
+
+// RoadmapSpec defines the desired state of Roadmap
+type RoadmapSpec struct {
```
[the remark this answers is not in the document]
julius-malcovsky: maybe see answer under comment "If a Roadmap references an ApiSpecification, which does not exist. How should this be handled?" 

im in between. both make sense ...

rover/internal/webhook/v1/roadmap_webhook.go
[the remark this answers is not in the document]
ron96g: Makes sense. It could be done using the priority ordering in rover-ctl, however, I agree that it should be an optional and lightweight documentation. There is the idea that we return warnings to Users, however, that has not been implemented yet and I think its out-of-scope. So lets keep it like it is right now

rover/api/v1/roadmap_types.go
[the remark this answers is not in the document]
ron96g: I just want to avoid adding a new way to reference resources, we often use explictly references via types.ObjectRef or types.TypedObjectRef but also implicitly via some internal convention. This now introduces a new way via ResourceType... Question is if it makes sense to use TypedObjectRef which (potentially) points to an object that does not exist yet. If we can accept that as a state then I would prefer that over ResourceType. ResourceName could then be called string Discriminator which either has the value of the basePath or the eventType.

rover-server/api/openapi.yaml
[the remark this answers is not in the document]
ron96g: In theory, it would also be possible to do some sort of mapping in rover-ctl. E. g. if provided file is `kind==ApiRoadmap`, rover-ctl maps this to the new structure. However, the idea of rover-ctl is to "keep it dumb". so if we do this, it must be really minimal.

ron96g:
Both ApiRoadmap and ApiChanlog are missing the handler implementation in https://github.com/telekom/controlplane/blob/main/rover-ctl/pkg/handlers/registry.go#L55

rover-server/api/openapi.yaml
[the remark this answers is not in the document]
julius-malcovsky: perhaps we could also deduce this from the request. for backwards compatibility - if there is no resourceName/resourceType -> must be an old resource = ApiChangelog. And new ones will be "proper". 

Lets talk about it

rover-server/internal/controller/roadmap.go
```diff
@@ -0,0 +1,479 @@
...
+
+func NewRoadmapController(stores *s.Stores) *RoadmapController {
+	return &RoadmapController{
+		stores: stores,
+		Store:  stores.RoadmapStore,
+	}
+}
+
+// CreateApiRoadmap implements the ServerInterface for creating roadmaps (POST)
+// Accepts the old Java API format with basePath
```
ron96g: "old Java API" mentioning this feels shitty

rover-server/internal/controller/roadmap.go
```diff
@@ -0,0 +1,479 @@
...
+func (r *RoadmapController) CreateApiRoadmap(c *fiber.Ctx) error {
```
ron96g: Why implement Create. It was not implemented by anything else

rover-server/internal/controller/roadmap.go
```diff
@@ -0,0 +1,479 @@
...
+	if err != nil {
+		return err
+	}
+
+	return c.Status(fiber.StatusAccepted).JSON(response)
+}
+
+// UpdateApiRoadmap implements the ServerInterface for updating roadmaps (PUT)
+// Accepts the old Java API format with basePath
+func (r *RoadmapController) UpdateApiRoadmap(c *fiber.Ctx, apiRoadmapId string) error {
```
ron96g: How do we ensure that apiRoadmapId actually is correct for the provided basePath? And why is it never used?

rover-server/internal/controller/roadmap.go
```diff
@@ -0,0 +1,479 @@
...
+		Name:        roadmapName,
+	}
+
+	// Upload to file-manager
+	fileAPIResp, err := r.uploadFile(ctx, itemsMarshaled, tempId)
+	if err != nil {
+		return res, err
+	}
+
+	// Construct TypedObjectRef to specification resource
```
ron96g: Why do the mapping here and not like all other controllers in a specific mapper pkg?

rover-server/internal/controller/roadmap.go
```diff
@@ -0,0 +1,479 @@
...
+		},
+		Spec: roverv1.RoadmapSpec{
+			SpecificationRef: apiSpecRef,
+			Contents:         fileAPIResp.FileId,
+			Hash:             fileAPIResp.FileHash,
+		},
+	}
+	EnsureLabelsOrDie(ctx, roadmap)
+
+	err = r.Store.CreateOrReplace(ctx, roadmap)
```
ron96g, line 97: I think we never use `CreateOrReplace` but instead `CreateOrUpdate` using a mutator. Did you check that this actually works as expected?

rover-server/internal/controller/roadmap.go
```diff
@@ -0,0 +1,479 @@
...
+	if err != nil {
+		return res, err
+	}
+
+	// Remove any duplicate roadmaps pointing to the same specification
+	if err := r.removeDuplicates(ctx, roadmap); err != nil {
```
ron96g: Is this `removeDuplicates` logic even needed in the new CP. It was added to the old one because at some point validation was disabled and user created duplicate resources

rover-server/internal/controller/roadmap.go
```diff
@@ -0,0 +1,479 @@
...
+		log.Warnf("Failed to remove duplicate roadmaps for basePath %s: %v", basePath, err)
+	}
+
+	// Download items to return in response
+	reader, err := r.downloadFile(ctx, fileAPIResp.FileId)
```
ron96g: Do we need to down it? We have just uploaded it

rover-server/internal/controller/roadmap.go
```diff
@@ -0,0 +1,479 @@
...
+
+	var responseItems []api.ApiRoadmapItem
+	err = json.NewDecoder(reader).Decode(&responseItems)
+	if err != nil {
+		return res, errors.Wrap(err, "failed to decode roadmap items")
+	}
+
+	// Construct response
+	resourceId := secCtx.Group + "--" + secCtx.Team + "--" + roadmapName
+	return api.ApiRoadmapResponse{
```
ron96g: mapper pkg

rover-server/internal/server/server.go
```diff
@@ -49,6 +49,14 @@ type EventSpecificationController interface {
 	GetStatus(ctx context.Context, resourceId string) (api.ResourceStatusResponse, error)
 }
 
+type ApiRoadmapController interface {
```
ron96g, line 52: This breaks the interface design. Should not pass fiber.Ctx to the Controller, see other controllers, e.g. `Update(ctx context.Context, resourceId string, req api.EventSpecificationUpdateRequest) (api.EventSpecificationResponse, error)`

rover-server/internal/server/roadmap_server.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"github.com/gofiber/fiber/v2"
+	"github.com/telekom/controlplane/common-server/pkg/problems"
+	"github.com/telekom/controlplane/common-server/pkg/server"
+
+	"github.com/telekom/controlplane/rover-server/internal/api"
+)
+
+func (s *Server) GetAllApiRoadmaps(c *fiber.Ctx) error {
```
ron96g, line 17: See https://github.com/telekom/controlplane/pull/302/changes#r3084898254

rover/internal/webhook/v1/roadmap_webhook.go
```diff
@@ -0,0 +1,80 @@
...
+	valErr := cerrors.NewValidationError(roverv1.GroupVersion.WithKind("Roadmap").GroupKind(), roadmap)
+
+	// Validate environment label is present
+	_, ok := controller.GetEnvironment(roadmap)
+	if !ok {
+		valErr.AddInvalidError(cerrors.MetadataEnvPath, "", "environment label is required")
+	}
+
+	// Validate specification reference is not empty
+	if roadmap.Spec.SpecificationRef.Name == "" {
```
ron96g: This is already validated using the CEL expressions

ron96g: Based on the validations done in this webhook. I dont think that we need this webhook

ron96g (changes requested):
- The webhook only check for non-empty properties which can be done using normal CEL via struct-annotations.
- There should be an explicit mapper pkg as was done for all other resources
- The Controller interface is invalid and should not contain the fiber.Ctx, see other controllers.
- I am not sure if removeDuplicates is still necessary as the validation MUST ensure that the normalized name matches the basePath and thus is unique.
- I have created a seperate issue to track the general issue of lossy basePath-conversion: https://github.com/telekom/controlplane/issues/335

ron96g (approved):

PR #307, opened 2026-03-31
test(spy-server): add tests & verify

stefan-ctrl:


ron96g (changes requested):
I cannot find any tests for the obfuscated secrets "******" (atleast not in the snapshosts). Could you check if this is working and add tests for it?

ron96g (approved):

PR #318, opened 2026-04-06
Feat/dynamic traffic control

julius-malcovsky:
DTC (Dynamic Traffic Control)

**Purpose:** Automatic cross-zone failover. Zones with DTC URLs are aggregated into a special "dtc" realm. When any subscription has failover enabled, routes use this realm so Kong can route traffic across all DTC-eligible zones.

**Key changes:**
- **`admin/zone_types.go`** — New optional `GatewayConfig.DtcUrl` field on zones.
- **`admin/zone/handler.go`** — `createGatewayDtcRealm()` builds a superset realm with the default gateway URL + all DTC URLs from all zones, paired with their issuers.
- **`route_util.go`** — `CreateProxyRoute()` now takes a `realmName` param. Downstream uses the provided realm (default or dtc); upstream always uses the default realm.
- **`apiexposure/handler.go`** — Selects DTC realm when `HasAnySubscriptionWithFailover()` returns true.
- **`rover/rover_types.go`** — New `FailoverEnabled` bool field. When true, failover zones are auto-discovered.
- **`rover/api/util.go`** — New `GetDtcEligibleZones()` returns zones with a DTC URL and Enterprise visibility.
- **`rover/api/subscription.go`** — When `FailoverEnabled`, auto-populates failover zones on ApiSubscription from DTC-eligible zones.


admin/internal/handler/zone/handler.go
```diff
@@ -181,7 +192,6 @@ func createTeamApiRoute(ctx context.Context, handlingContext HandlingContext, te
 
 	mutator := func() error {
 		teamRoute.Labels = map[string]string{
```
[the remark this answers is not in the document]
julius-malcovsky: the env label is automatically added by the scoped client were using, so explicit is not required

admin/internal/handler/util/naming/naming.go
```diff
@@ -30,6 +30,10 @@ func ForTeamApiGatewayRealm(environment *adminv1.Environment) string {
 	return teamApiIdentityRealmPrefix + environment.GetName()
 }
 
+func ForDtcGatewayRealm() string {
+	return "dtc"
+}
+
```
stefan-ctrl, line 39: A comment above `ForDtcGatewayRealm` be nice, explaining why this value is being hardcoded wand what it is used for (i.e. explain the convetion).

Also, "dtc"  could also be a constant.

api/internal/handler/apiexposure/handler.go
```diff
@@ -86,38 +86,99 @@ func (h *ApiExposureHandler) CreateOrUpdate(ctx context.Context, apiExp *apiapi.
...
-		return errors.Wrapf(err, "failed to check cross-zone subscribers for apiExposure: %s", apiExp.Name)
+		return errors.Wrapf(err, "failed to check for failover subscriptions for apiExposure: %s", apiExp.Name)
 	}
 
-	route, err := util.CreateRealRoute(ctx, apiExp.Spec.Zone, apiExp, contextutil.EnvFromContextOrDie(ctx),
-		util.WithProxyTarget(hasCrossZoneSubs),
-	)
+	realmName := contextutil.EnvFromContextOrDie(ctx) // default
+	if hasFailoverSubscription {
+		realmName = "dtc"
```
stefan-ctrl, line 102: Perhabs in `common` or `admin/api` why should have a constant for that, that would be imported here.

gateway/api/v1/realm_types_test.go
```diff
@@ -0,0 +1,185 @@
...
+			Expect(downstreams).To(HaveLen(4))
+
+			// First two downstreams should have matching issuers
+			Expect(downstreams[0].IssuerUrl).To(Equal("https://idp1.example.com/auth/realms/dtc"))
+			Expect(downstreams[1].IssuerUrl).To(Equal("https://idp2.example.com/auth/realms/dtc"))
+
+			// Last two downstreams should have empty issuer URLs
+			Expect(downstreams[2].IssuerUrl).To(BeEmpty())
+			Expect(downstreams[3].IssuerUrl).To(BeEmpty())
+		})
```
stefan-ctrl, line 111: What is the scenario for such a mismatch in Downstreams and IssuerUrls? When do we expect this to happen?

stefan-ctrl (commented):
Great Job 👍 

Have a simple question regarding Empty-Issuers in Downstream configuration and a small remark regarding harded-strings.

gateway/api/v1/realm_types_test.go
julius-malcovsky: this can happen with DTC configurations. if you have 3 zones with DTC, then you have 3 + 1 urls (dtc of each zone + the current zones own non-dtc url), but there are no dtc issuers. so you will only have the 3 zones issuers. thus you have 4 vs 3. 

also, please be aware that we want to refactor this and move the issuer(s) outside of the downstreams. most likely as a field in security. this will make things clearer, remove the n:m confusion and remove the need to deduplicate values

admin/internal/handler/util/naming/naming.go
julius-malcovsky: added a constant + comment

api/internal/handler/apiexposure/handler.go
julius-malcovsky: here i just added a comment. importing was problematic due to circles and i think putting it into common would work, but feels like an overkill scope - feel free to discuss it

PR #320, opened 2026-04-09
Feat/Controlplane API Write Operations

iagarba:


.github/scripts/tidy-all-modules.sh
ron96g, line 1: Do not change the filemode. Same for the other files.

application/internal/webhook/v1/mutator/mutate.go
```diff
@@ -0,0 +1,57 @@
...
+	log := logr.FromContextOrDiscard(ctx)
+
+	if secretsapi.IsRef(app.Spec.Secret) {
+		log.V(1).Info("spec.secret is already a reference, nothing to do")
+		return nil
+	}
+
+	var clientSecret string
+	if strings.EqualFold(app.Spec.Secret, secret.KeywordRotate) || app.Spec.Secret == "" {
+		clientSecret = string(uuid.NewUUID())
```
ron96g: Should use https://github.com/telekom/controlplane/blob/main/secret-manager/api/secretgen.go#L27

controlplane-api/.claude/settings.local.json
ron96g, line 1: Local dev file. Do not commit

controlplane-api/cmd/main.go
```diff
@@ -58,7 +69,24 @@ func main() {
...
+		}
+		teamService = service.NewTeamK8sService(k8sClient)
+		applicationService = service.NewApplicationK8sService(k8sClient)
+		approvalService = service.NewApprovalK8sService(k8sClient)
+		log.Info("Kubernetes integration enabled")
+	} else {
+		log.Info("Kubernetes integration disabled, mutations will be unavailable")
+	}
+
+	srv := newGraphQLServer(client, teamService, applicationService, approvalService, cfg.Security.Enabled)
```
ron96g: Could put all services into a temporary struct to reduce number of parameters

controlplane-api/internal/resolvers/model/mutation_types.go
ron96g, line 1: Was this file generated?

controlplane-api/internal/resolvers/model/mutation_types.go
```diff
@@ -0,0 +1,103 @@
+// Copyright 2025 Deutsche Telekom IT GmbH
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package model
```
ron96g, line 5: In these structures we should think "What does the user or UI need to continue?", e.g. it will probably follow-up with a query of the created/updated team so we should include that information

controlplane-api/internal/resolvers/model/mutation_types.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// CreateTeamInput is the input for creating a new team.
+type CreateTeamInput struct {
+	Environment string            `json:"environment"`
+	Group       string            `json:"group"`
+	Name        string            `json:"name"`
+	Email       string            `json:"email"`
+	Members     []MemberInput     `json:"members"`
+	Category    TeamCategoryInput `json:"category"`
```
ron96g: At the momemt Category cannot be configured, its always Customer

controlplane-api/internal/resolvers/model/mutation_types.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// UpdateTeamInput is the input for updating an existing team.
+type UpdateTeamInput struct {
```
ron96g, line 25: Does it make sense to add a NodeId in here? Which is the primary-key for all resources in the GraphQL-API? So "update team with ID 42"

controlplane-api/internal/resolvers/model/mutation_types.go
```diff
@@ -0,0 +1,103 @@
...
+
+// RotateApplicationSecretInput is the input for rotating an application's client secret.
+type RotateApplicationSecretInput struct {
+	Environment string `json:"environment"`
+	Team        string `json:"team"`
+	Name        string `json:"name"`
+}
+
+// ApplicationMutationResult is the response type for application mutations.
+type ApplicationMutationResult struct {
```
ron96g: Why not call it `RotateApplicationSecretResult`? And can be directly return the new clientSecret here?

controlplane-api/internal/resolvers/model/mutation_types.go
```diff
@@ -0,0 +1,103 @@
...
+type RotateTeamTokenInput struct {
+	Environment string `json:"environment"`
+	Group       string `json:"group"`
+	Name        string `json:"name"`
+}
+
+// DecisionInput represents the decision details for approval mutations.
+type DecisionInput struct {
+	Name    string  `json:"name"`
+	Email   *string `json:"email,omitempty"`
```
ron96g: We must require email. And this will be difficult to manage, because this should not be a free-text field but instead we should extract the email from the current logged-in user

controlplane-api/internal/resolvers/model/mutation_types.go
```diff
@@ -0,0 +1,103 @@
...
+// RotateTeamTokenInput is the input for rotating a team's token.
+type RotateTeamTokenInput struct {
+	Environment string `json:"environment"`
+	Group       string `json:"group"`
+	Name        string `json:"name"`
+}
+
+// DecisionInput represents the decision details for approval mutations.
+type DecisionInput struct {
+	Name    string  `json:"name"`
```
ron96g, line 63: Not sure about Name here

controlplane-api/internal/resolvers/model/mutation_types.go
```diff
@@ -0,0 +1,103 @@
...
+type ApprovalMutationResult struct {
+	Success      bool    `json:"success"`
+	Message      string  `json:"message"`
+	NewState     *string `json:"newState,omitempty"`
+	Namespace    *string `json:"namespace,omitempty"`
+	ResourceName *string `json:"resourceName,omitempty"`
+}
+
+// TeamCategoryInput represents the team category enum for mutations.
+type TeamCategoryInput string
```
ron96g: Delete

controlplane-api/internal/resolvers/model/mutation_types.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// DecisionInput represents the decision details for approval mutations.
+type DecisionInput struct {
+	Name    string  `json:"name"`
+	Email   *string `json:"email,omitempty"`
+	Comment *string `json:"comment,omitempty"`
+}
+
+// DecideApprovalRequestInput is the input for deciding on an ApprovalRequest.
```
ron96g, line 68: Same as the other Inputs and Results: Should we add the nodeId in here?

controlplane-api/internal/service/application_k8s.go
```diff
@@ -0,0 +1,61 @@
...
+	scopedClient := cc.NewScopedClient(s.client, input.Environment)
+
+	// Fetch existing Application to verify it exists
+	app := &applicationv1.Application{}
+	err := scopedClient.Get(ctx, client.ObjectKey{Namespace: namespace, Name: resourceName}, app)
+	if err != nil {
+		return nil, mapK8sError(err)
+	}
+
+	_, err = scopedClient.CreateOrUpdate(ctx, app, func() error {
```
ron96g, line 47: Question: Does it make sense to use Patch(...) instead of CreateOrUpdate(...)? I think its more efficient because its applied on the K8s-Server and we just set clientSecret attribute

controlplane-api/internal/service/approval.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"context"
+
+	"github.com/telekom/controlplane/controlplane-api/internal/resolvers/model"
+)
+
+// ApprovalService defines operations for managing Approval and ApprovalRequest resources.
+type ApprovalService interface {
+	DecideApprovalRequest(ctx context.Context, input model.DecideApprovalRequestInput) (*model.ApprovalMutationResult, error)
```
ron96g, line 15: [nitpick] Personally I like to call it TranisitonXYZ as it fits more into the FSM pattern

controlplane-api/internal/service/application.go
ron96g, line 1: Why split it into xyz.go and xyz_k8s.go?

controlplane-api/internal/service/approval_k8s.go
```diff
@@ -0,0 +1,172 @@
...
+	return &approvalK8sService{client: c}
+}
+
+// approvalNamespace returns the K8s namespace for approval resources: <environment>--<team>.
+func approvalNamespace(environment, team string) string {
+	return environment + "--" + team
+}
+
+// mapActionToApprovalAction maps a GraphQL action string (uppercase) to the CRD ApprovalAction (title-case).
+func mapActionToApprovalAction(action string) (approvalv1.ApprovalAction, error) {
```
ron96g, line 34: Question: Why did we decide to make it all upper-case in the API and internally different? Why not just use the internal format and get rid of all these enum mappings?

controlplane-api/internal/service/approval_k8s.go
```diff
@@ -0,0 +1,172 @@
...
+		return nil, err
+	}
+
+	targetState, err := findTargetState(ar.Status.AvailableTransitions, crdAction)
+	if err != nil {
+		return nil, err
+	}
+
+	// Update the CRD
+	_, err = scopedClient.CreateOrUpdate(ctx, ar, func() error {
```
ron96g, line 104: Same as the other comment: Patch vs CreateOrUpdate

controlplane-api/internal/service/helpers.go
```diff
@@ -0,0 +1,126 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/telekom/controlplane/controlplane-api/internal/viewer"
+	"github.com/vektah/gqlparser/v2/gqlerror"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+)
+
+// authorizeCreateTeam checks that the viewer is allowed to create a team in the given group.
```
ron96g, line 16: Should these function be here or at the central place where all other authorize function are located?

controlplane-api/internal/service/helpers.go
```diff
@@ -0,0 +1,126 @@
...
+	return group + "--" + team
+}
+
+// mapK8sError converts a Kubernetes API error to a GraphQL error with an appropriate error code.
+func mapK8sError(err error) *gqlerror.Error {
+	if err == nil {
+		return nil
+	}
+
+	switch {
```
ron96g, line 94: Could be more generic by converting to K8s statusError or make it more specific by "Failed to rotate your APplication secret because ...". I think, this does not add any real benefit

install/local/controlplane-api-config.yaml
ron96g, line 1: Should decide on config and not update for local testing

install/local/controlplane-api-config.yaml
```diff
@@ -16,3 +18,7 @@ graphql:
 
 log:
   level: "info"
+
+kubernetes:
+  enabled: true
+  kubeconfig: "/home/igar/.kube/config"  # path to your kubeconfig
```
ron96g: personal path

install.sh
ron96g, line 1: no filemode changes

ron96g (changes requested):
Looks good. A few minor issues and some more general question we need to decide on

controlplane-api/internal/resolvers/model/mutation_types.go
iagarba: yes, it was generated :)

controlplane-api/internal/resolvers/model/mutation_types.go
iagarba: I decided against it, but let's talk about it if it makes sense to do that

controlplane-api/internal/resolvers/model/mutation_types.go
iagarba: What do you mean?

controlplane-api/internal/resolvers/model/mutation_types.go
iagarba: Noted!

PR #321, opened 2026-04-10
feat: added new projector domain; updated controlplane-api

ron96g:
- **feat(controlplane-api): evolve schemas for projector and relocate shared model types**
- **feat(projector): add CQRS projector for materializing CR state into PostgreSQL**


projector/internal/bootstrap/bootstrap.go
```diff
@@ -0,0 +1,201 @@
...
+	"github.com/telekom/controlplane/projector/internal/infrastructure"
+	"github.com/telekom/controlplane/projector/internal/module"
+)
+
+var (
+	scheme   = runtime.NewScheme()
+	setupLog = ctrl.Log.WithName("setup")
+)
+
+func init() {
```
BjoernKarma, line 55: Would it be helpful if we logged errors during the scheme registration?

controlplane-api/ent/schema/application.go
```diff
@@ -39,6 +39,10 @@ func (Application) Fields() []ent.Field {
 			Optional().
 			Nillable().
 			NotEmpty(),
+		field.Text("client_secret").
+			Optional().
+			Nillable().
+			NotEmpty(),
```
BjoernKarma, line 45: Does it make sense to add the sensitive() annotation to exclude from logging/serialization  (https://entgo.io/docs/schema-fields/#sensitive-fields).

projector/Makefile
```diff
@@ -0,0 +1,28 @@
+# Copyright 2026 Deutsche Telekom IT GmbH
+#
+# SPDX-License-Identifier: Apache-2.0
+
+BUILD_TIME=$(shell date -u +"%Y-%m-%dT%H:%M:%SZ")
+VERSION=$(shell git describe --tags --always --dirty || echo 'develop')
+GIT_COMMIT=$(shell git rev-parse HEAD)
+
```
BjoernKarma, line 8: If we want, we could harmonize the Makefile (e.g. with controlplane-api) and add targets for: lint, lint-fix and clean

projector/README.md
```diff
@@ -0,0 +1,119 @@
...
+- **Command side** — Kubernetes Custom Resources are the source of truth; operators and users mutate them through the K8s API.
+- **Query side** — The projector continuously watches those resources and materialises a relational read model in PostgreSQL, optimised for the GraphQL query layer.
+
+The projector bridges these two worlds: it subscribes to change events on the command side and maintains a **materialised projection** — a denormalised, query-friendly representation of the same data. This is analogous to a materialised view in event-sourcing architectures, except the "event log" is the Kubernetes watch stream.
+
+## Architecture
+
+```
+main.go                          26-line entrypoint → bootstrap.Run()
+│
```
BjoernKarma, line 24: Do we want to maintain this folder structure and keep it in sync with the filesystem? I like the explanation and overview it gives, but it quickly gets stale. Or maybe we could add a hint saying this only contains the most important folders and is not complete?

BjoernKarma (approved):
Looks very good. I like the structure and the design and also the way the abstraction works with the contracts. Great work!

I only have a few minor remarks added as comments. Maybe we can also in addition have a look at:
- There are some minor golangci-lint issues that might be fixed
- The component has `run_check_generated_files` disabled (same as controlplane-api). Maybe in the future it makes sense to also check the generated files, but this time not for kubebuilder but ent generated files
- We need to keep in mind that in the future audit logs and soft-delete with history are requirements that could become relevant. Not something we need to implement right now, just make sure it is possible in the future

projector/internal/bootstrap/bootstrap.go
ron96g: We should wrap it in a `Must(...)` function which panics as this should never happen. Will change

controlplane-api/ent/schema/application.go
ron96g: The problem is that it will then also be omitted from encoding and then it can never be returned. Under normal deployment, this value always contains the secret-reference anyways, which is non-sensitive

projector/Makefile
ron96g: Harmonizing the Makefiles in general is a great idea. But maybe in a different PR?

projector/README.md
ron96g: I agree. Lets keep it high-level to increase maintainability. I will change that

PR #323, opened 2026-04-13
feat(identity): refactor keycloak integration and add graceful client-secret rotation

ron96g:
## Summary

- **Refactor** the identity module's Keycloak integration layer: replace the monolithic `RealmClient` with a `KeycloakService` interface and SHA-256-keyed `ServiceFactory` for per-realm client caching; extract mappers into a reusable `util/` package
- **Add graceful client-secret rotation** as a new feature: `SecretRotationConfig` on `RealmSpec` generates a new secret, configures a Keycloak client-policy to accept both old and new secrets during a transition window, and exposes `RotatedClientSecret` / `RotatedSecretExpiresAt` on `ClientStatus`
- **Fix multiple bugs** found during code review: realm status persistence regression, nil-pointer dereferences in service responses, secret leakage via default-level logs, non-idempotent `DeleteRealm`, stale rotation status fields, and dead code

### Refactoring
- Replace `RealmClient` interface with `KeycloakService` + `ServiceFactory` (dependency injection)
- Rename `mapper/` → `util/` with nil-safe generic helpers (`ptrEqual[T]`)
- Delete legacy `client.go`, `client_factory.go`, `client_implementation.go`, `util.go`
- Migrate `common/` error handling from `pkg/errors` to stdlib `errors.As`
- Remove Grafana dashboard JSON files
- Update OpenAPI spec and regenerate client code

### New Feature — Graceful Secret Rotation
- `SecretRotationConfig` field on `RealmSpec` (CRD addition)
- `graceful_secrets.go` — rotation lifecycle: generate → policy → transition → expire
- `RotatedClientSecret` and `RotatedSecretExpiresAt` on `ClientStatus` (CRD addition)
- `protocolmappers/client_id.go` — extracted client-ID protocol mapper

### Bug Fixes (from code review)
- **P0-1**: Persist `realm.Status` back to CR after handler execution
- **P0-2**: Redact HTTP response bodies from default-level log output
- **P0-3**: Nil-pointer guard after `CreateClient` when `resBody.Id` is nil
- **P1-1**: Clear `RotatedClientSecret`/`RotatedSecretExpiresAt` when secret is a reference
- **P1-2**: Nil guard on `refreshed.JSON2XX` after post-rotation re-fetch
- **P1-3**: Accept HTTP 404 as idempotent success in `DeleteRealm`
- **P1-4**: Fix misleading error message (`"failed to get realm clients"` → `"failed to delete client"`)
- **P2-1**: Delete dead code (`util.go` — 4 unused functions)
- **P2-2**: Fix errcheck and unnecessary type conversions (lint)
- **P2-3**: Add nil guards in `util/client.go`, `util/realm.go`, `util/protocolmapper.go`

## Testing

- **58/58 identity tests pass** (19 client + 2 IdP + 24 realm + 13 keycloak pkg)
- Lint clean on all modified packages
- Added 6 realm status persistence assertions and updated 4 client handler test fixtures

## Stats

80 files changed, +6171 / −6781 lines


identity/pkg/keycloak/keycloak_service.go
```diff
@@ -0,0 +1,331 @@
...
+// representation. When differences are detected the existing object is merged
+// with the desired fields (preserving Keycloak-managed attributes such as Id,
+// DefaultClientScopes, Attributes, etc.) and PUT back. If the client secret
+// changed and the realm supports graceful rotation, a forced secret rotation is
+// triggered first so that Keycloak's rotation executor preserves the old secret
+// in the "rotated" slot.
+func (k *keycloakService) updateClient(ctx context.Context, realmName string, existing *api.ClientRepresentation, client *identityv1.Client, supportsGracefulRotation bool) error {
+	logger := logr.FromContextOrDiscard(ctx)
+	clientUUID := *existing.Id
+	client.Status.ClientUid = clientUUID
```
lukas016, line 226: why is clientuid and not ClientUUID

identity/pkg/keycloak/keycloak_service.go
```diff
@@ -0,0 +1,331 @@
...
+
+	// Merge desired fields into the existing representation and PUT.
+	merged := util.MergeClientRepresentation(existing, &desired)
+	body := *merged
+
+	res, err := k.Client.PutRealmClientsIdWithResponse(ctx, realmName, clientUUID, body)
+	if err != nil {
+		return fmt.Errorf("error updating client: %w", err)
+	}
+	if res.StatusCode() != 204 && res.StatusCode() != 200 {
```
lukas016: optional: 200 and 204 are technically magic number. Recommendation is used http library and constant with name which adding some semantic information.

common/pkg/errors/ctrlerrors/errors.go
```diff
@@ -35,36 +35,39 @@ type RetryableWithDelayError interface {
...
 // that suggests whether to requeue the reconciliation and after what duration.
 func HandleError(ctx context.Context, obj types.Object, err error, recorder record.EventRecorder) (bool, reconcile.Result) {
-	rootCauseErr := errors.Cause(err)
-
-	if be, ok := rootCauseErr.(BlockedError); ok && be.IsBlocked() {
-		recordError(ctx, obj, rootCauseErr, "Blocked", recorder)
-		updatd := obj.SetCondition(condition.NewBlockedCondition(rootCauseErr.Error()))
-		return updatd, reconcile.Result{
+	var be BlockedError
+	if errors.As(err, &be) && be.IsBlocked() {
```
lukas016, line 44: optional: if we use golang 1.26, it is more safe to use: https://pkg.go.dev/errors#example-AsType

identity/pkg/keycloak/keycloak_service.go
ron96g: Not sure, probably just oversight. But minor cosmetic, will not change

identity/pkg/keycloak/keycloak_service.go
ron96g: That is true and for the future, I will keep this in mind, however, will not change now.

common/pkg/errors/ctrlerrors/errors.go
ron96g: Nice feature. But this should be refactored in general. Right now we are using errors.As(...) everywhere else

identity/internal/handler/realm/handler.go
```diff
@@ -34,48 +41,52 @@ func (h *HandlerRealm) CreateOrUpdate(ctx context.Context, realm *identityv1.Rea
...
-			return nil
+			return ctrlerrors.BlockedErrorf("IdentityProvider %q not found", realm.Spec.IdentityProvider.String())
 		}
 		return err
 	}
-	idpSpec := identityprovider.ObfuscateIdentityProvider(identityProvider.Spec)
-	logger.V(0).Info("Found IdentityProvider", "idp", idpSpec)
 
-	var realmStatus = MapToRealmStatus(identityProvider, realm.Name)
+	var realmStatus = mapToRealmStatus(identityProvider, realm.Name)
```
BjoernKarma, line 49: GetIdentityProviderByName returns (nil, nil) when the IDP exists but its Ready condition is false. The code has no nil check before calling mapToRealmStatus(identityProvider, realm.Name) on line 50. Should we add a nil guard after line 47?

identity/pkg/keycloak/graceful_secrets.go
```diff
@@ -0,0 +1,388 @@
...
+	} else {
+		log.V(1).Info("existing rotated secret deleted or not found, proceeding with rotation", "realm", realmName, "clientId", keycloakId)
+	}
+
+	res, err := k.Client.PostRealmClientsIdClientSecretWithResponse(ctx, realmName, keycloakId)
+	if err != nil {
+		return fmt.Errorf("error forcing secret rotation: %w", err)
+	}
+	if res.StatusCode() != 200 {
+		return fmt.Errorf("unexpected status %d when forcing secret rotation", res.StatusCode())
```
BjoernKarma: This uses a raw res.StatusCode() != 200 check and returns a plain fmt.Errorf instead of using CheckHTTPStatus/CheckStatusCode. This means the error won't be classified as blocked vs. retryable.  Should we use  CheckHTTPStatus here?

identity/pkg/keycloak/keycloak_service.go
```diff
@@ -0,0 +1,350 @@
...
+
+	log := logr.FromContextOrDiscard(ctx)
+
+	if len(clientUid) > 0 {
+		clientRes, err := k.Client.GetRealmClientsIdWithResponse(ctx, realmName, clientUid)
+		if err != nil {
+			return nil, fmt.Errorf("unexpected error when fetching client by UID: %w", err)
+		}
+		if clientRes.StatusCode() == 200 && clientRes.JSON2XX != nil {
+			log.V(2).Info("client found by uuid", "response", string(clientRes.Body))
```
BjoernKarma: Is it possible that the client response contains sensitive data that would be logged? Should we maybe only include clientID and statusCode in the log?

identity/internal/handler/realm/handler.go
BjoernKarma: Is it possible that the realm contains sensitive data that would be logged? Should we maybe only include name and namespace of the realm in the log?

identity/pkg/keycloak/keycloak_service.go
```diff
@@ -0,0 +1,350 @@
...
+	if util.CompareClientRepresentation(existing, &desired) {
+		logger.V(1).Info("no changes detected for client, skipping update",
+			"clientId", client.Spec.ClientId, "keycloakId", clientUUID)
+		return nil
+	}
+
+	// If the secret changed and the realm supports graceful rotation, force
+	// rotation before PUT so Keycloak moves the current secret into the
+	// "rotated" slot with the configured grace period.
+	if util.HasSecretChanged(existing, &desired) && supportsGracefulRotation {
```
BjoernKarma: What happens if the PUT fails and is then re-tried? Is it idempotent or is the original secret (that should we rotated in the first place) lost?

identity/pkg/api/client.go
```diff
@@ -1,35 +0,0 @@
-// Copyright 2025 Deutsche Telekom IT GmbH
```
BjoernKarma, line 1: If I remember correctly, the reason for having this API client was to mock only parts of the generated interface for tests. Then, the tests can cover all the code that is not generated. Do we keep this in the refactored structure as well? That means, do we have the same test coverage in the new structure?

BjoernKarma (commented):
Looks very good. I have added a few questions as comments. Maybe we can have a look?

identity/api/v1/realm_types.go
```diff
@@ -75,3 +119,15 @@ func (el *RealmList) GetItems() []types.Object {
 func init() {
 	SchemeBuilder.Register(&Realm{}, &RealmList{})
 }
+
+var SecretRotationConditionType = "SecretRotation"
+
+func NewSecretRotatedCondition(rotatedAt time.Time) metav1.Condition {
```
julius-malcovsky: cosmetic: i think we have a conditions.go where we place funcs like this right? I understand its a single func, but might be worth it to keep things consistent.

identity/pkg/keycloak/client_error.go
```diff
@@ -5,49 +5,113 @@
...
 
+// ApiError represents an error from the Keycloak API.
+// It implements the ctrlerrors.BlockedError, ctrlerrors.RetryableError, and
+// ctrlerrors.RetryableWithDelayError interfaces via duck-typing, so that
+// errors propagated up the call stack are correctly classified by
+// ctrlerrors.HandleError without introducing a direct import dependency.
 type ApiError interface {
 	error
 	Retriable() bool
+	IsRetryable() bool
```
julius-malcovsky, line 27: whats the difference between 

Retriable() bool **and** IsRetryable() bool

identity/internal/handler/realm/handler.go
```diff
@@ -34,48 +41,52 @@ func (h *HandlerRealm) CreateOrUpdate(ctx context.Context, realm *identityv1.Rea
...
+		return ctrlerrors.BlockedErrorf("IdentityProvider %q is not valid: %s", realm.Spec.IdentityProvider.String(), err)
 	}
 
 	// Create a copy of the realmStatus so that we NEVER modify the original status
 	// and accidentally write the secrets back to the cluster
 	replacedRealmStatus := realmStatus.DeepCopy()
 	replacedRealmStatus.AdminPassword, err = secrets.Get(ctx, realmStatus.AdminPassword)
 	if err != nil {
-		return errors.Wrap(err, "failed to retrieve password from secret manager")
+		return fmt.Errorf("failed to retrieve password from secret manager: %w", err)
```
julius-malcovsky, line 60: any particular reason ?

julius-malcovsky (dismissed):
mostly cosmetic, otherwise LGTM

identity/pkg/api/client.go
ron96g: This has been moved to keycloak_service.go KeycloakClient interface so that the module itself describes its required client interface.

identity/pkg/keycloak/keycloak_service.go
ron96g: Thats a good catch. Currently this would break the graceful rotation and the original, rotated secret would be lost. Will investigate how to fix

identity/internal/handler/realm/handler.go
ron96g: The realm should only contain sensitive data if the resolved adminSecret has been stored in it. However, will refactor anyways (too spammy in the logs)

identity/pkg/keycloak/keycloak_service.go
ron96g: Yes, this would contain the unencoded secret but only on debug-level. However, will refactor

identity/pkg/keycloak/graceful_secrets.go
ron96g: Good catch

identity/internal/handler/realm/handler.go
ron96g: Will refactor to use new `condition.EnsureReady(...)` logic

identity/internal/handler/realm/handler.go
ron96g: the pkg/errors that we used for `errors.Wrap` is not longer maintained :( The best practice is `fmt.Errorf(...)`

See https://go.dev/blog/go1.13-errors#wrapping-errors-with-w

So ideally we use it in all newly written code from now on

identity/pkg/keycloak/client_error.go
ron96g: `IsRetryable()` implements our common and common-server Error interfaces while `Retriable()` was there before and kept for compatibility

identity/api/v1/realm_types.go
ron96g: Yes, lets separate it

identity/internal/handler/realm/handler.go
lukas016: it isn't maintained more as 2 years already 😄

identity/pkg/keycloak/token_source.go
```diff
@@ -0,0 +1,101 @@
...
+
+	token, err := s.inner.Token()
+	if err == nil {
+		return token, nil
+	}
+
+	// Only re-authenticate when the refresh token is actually invalid
+	// (400/401). For transient errors (network, 5xx, 429) propagate a
+	// classified error so the controller can apply correct retry logic.
+	if !isTokenInvalid(err) {
```
lukas016, line 56: optional: this is technically double negation. you can define isTokenValid instead of it.

identity/pkg/keycloak/token_source.go
```diff
@@ -0,0 +1,101 @@
...
+		cfg:      cfg,
+		ctx:      ctx,
+		username: username,
+		password: password,
+		inner:    cfg.TokenSource(ctx, initial),
+	}
+}
+
+func (s *reAuthTokenSource) Token() (*oauth2.Token, error) {
+	s.mu.Lock()
```
lukas016, line 45: optional: critical section is very long. in most cases you just need protect lines after 60 line. Plus probably you can use readwrite mutex to optimize performance, if you have more reads as write operations, if not you can keep mutex and try optimize it with TryLock.

identity/pkg/keycloak/keycloak_service.go
```diff
@@ -0,0 +1,366 @@
...
+
+func (k *keycloakService) createClient(ctx context.Context, realmName string, client *identityv1.Client) error {
+	logger := logr.FromContextOrDiscard(ctx)
+
+	body := util.MapToClientRepresentation(client)
+	res, err := k.Client.PostRealmClientsWithResponse(ctx, realmName, body)
+	if err != nil {
+		return fmt.Errorf("error creating client: %w", err)
+	}
+	if responseErr := CheckHTTPStatus(res.StatusCode(), 201); responseErr != nil {
```
lukas016, line 182: optional: magic number 201. we should use http.Status code instead of number.

identity/pkg/keycloak/util/client.go
```diff
@@ -0,0 +1,136 @@
...
+// compareSecretRotationAttribute returns true when both representations agree
+// on the controlplane.secret-rotation attribute value. The attribute is
+// considered "set" only when its value is the string "true".
+func compareSecretRotationAttribute(a, b *api.ClientRepresentation) bool {
+	return getSecretRotationAttr(a) == getSecretRotationAttr(b)
+}
+
+// getSecretRotationAttr returns "true" when the secret-rotation attribute is
+// present and set to "true", otherwise "".
+func getSecretRotationAttr(c *api.ClientRepresentation) string {
```
lukas016, line 88: question: why cannot we return true or false directly?

lukas016:
Main logic for rotation of secret LGTM but it should be tested before deploying on production.

BjoernKarma (approved):
LGTM - as discussed

PR #338, opened 2026-04-15
feat: improved projector

ron96g:
- **fix: moved schema migration to projector to avoid conflicts**
- **feat: added member indexes**
- **fix(projector): improved team member upsert handling using new indexes**
- **ci(projector): added gorealeser config**


controlplane-api/ent/migrate/schema.go
```diff
@@ -260,6 +260,18 @@ var (
 				OnDelete:   schema.SetNull,
 			},
 		},
+		Indexes: []*schema.Index{
+			{
```
stefan-ctrl, line 264: Why does this index has to be unique while "member_email" isn't?

projector/internal/domain/team/repository.go
```diff
@@ -173,19 +173,26 @@ func (r *Repository) withTx(ctx context.Context, fn func(tx *ent.Tx) error) erro
...
 	for _, m := range members {
-		memberEntity, err := tx.Member.Create().
+		id, err := tx.Member.Create().
 			SetName(m.Name).
 			SetEmail(m.Email).
 			SetTeamID(teamID).
-			Save(ctx)
+			OnConflictColumns(member.FieldEmail, member.TeamColumn).
+			Update(func(u *ent.MemberUpsert) {
+				u.SetName(m.Name)
```
stefan-ctrl, line 189: Doesn't email need to be set here as well, or is this not the case because of indices and primary key constraints? (i.e., if mail changes, member is rather treated as a new member instead of updating)

controlplane-api/ent/migrate/schema.go
ron96g: email cannot be unique as a member with email "x@y.com" can be part of multiple teams. if email would be unique, this would not be possible. The composite PK can be unique because there can only be one user with email "x@y.com" per team

projector/internal/domain/team/repository.go
ron96g: email is the identifier of a member so we cannot update it because if it changed, we dont have any clue what to update... If a user was removed (aka email changed) then this will be picked up by the garbage collection.

"if mail changes, member is rather treated as a new member instead of updating)" --> Yes

stefan-ctrl (approved):

PR #341, opened 2026-04-16
feat: make controlplane-api and projector optional in install overlays

ron96g:


install/overlays/default/kustomization.yaml
```diff
@@ -24,9 +24,14 @@ resources:
   - https://github.com/telekom/controlplane//file-manager/config/default/?timeout=120&ref=v0.18.0
   - https://github.com/telekom/controlplane//notification/config/default/?timeout=120&ref=v0.18.0
 
-# Uncomment to enable the eventing subsystem (event + pubsub controllers):
+# Uncomment to enable optional components:
 # components:
 #   - ../../components/eventing
+#   - ../../components/database
+
+# Uncomment to enable the controlplane-api and projector:
```
BjoernKarma: If this block is uncommented together with components, then it is outside the resources block and the indentation is 2-space.
 
- Should we move this directly under the resources block (line 27)? 
- Should we move the components block (currently line 27-30) after this?
- Should we adjust indentation, so that after uncommenting the indentation is correct?

install/overlays/local/controlplane-api-config.yaml
```diff
@@ -5,11 +5,6 @@
 database:
   url: ${DATABASE_URL}
 
-server:
-  address: ":8080"
```
BjoernKarma, line 9: Should we keep these overrides for the local overlay? Otherwise a local deployment requires cert-manager and might fail.

install/components/database/kustomization.yaml
```diff
@@ -0,0 +1,25 @@
...
+resources:
+  - postgres.yaml
+
+secretGenerator:
+  - name: controlplane-db
+    options:
+      disableNameSuffixHash: true
+    literals:
+      - url=postgres://controlplane:controlplane@controlplane-postgres:5432/controlplane?sslmode=disable
+      - username=controlplane
```
BjoernKarma, line 27: Should we add a more prominent warning here that this is only for testing and not for prod? Just to make sure that the literals are not forgotten to be modified for other environments.

install/overlays/default/kustomization.yaml
```diff
@@ -24,9 +24,14 @@ resources:
   - https://github.com/telekom/controlplane//file-manager/config/default/?timeout=120&ref=v0.18.0
   - https://github.com/telekom/controlplane//notification/config/default/?timeout=120&ref=v0.18.0
 
-# Uncomment to enable the eventing subsystem (event + pubsub controllers):
+# Uncomment to enable optional components:
```
BjoernKarma, line 30: Should we add a comment that explains that the database component is mandatory? Either by deploying the optional component or providing a valid URL with credentials (external database secret named controlplane-db)?

BjoernKarma (dismissed):
I like the changed flow for the install and the overlays. This makes it easier to use and also decide if optional components should be installed. I just have 4 small points. Maybe you can have a look at them?

install/overlays/default/kustomization.yaml
ron96g: I would do that in the docs. There we should have a dedicated section about the controlplane-api

install/overlays/default/kustomization.yaml
ron96g: yes, this needs to be moved after line 25

install/overlays/local/controlplane-api-config.yaml
ron96g: "local" refers to the deployment to a kind cluster which - if setup correctly using scripts - has the cert-manager installed. This is the same for all others components as well.

install/components/database/kustomization.yaml
ron96g: Good idea

BjoernKarma (approved):
LGTM

PR #347, opened 2026-04-20
ci: Add coverage and test reports in our workflow

BjoernKarma:
- Replace gotestfmt with gotestsum across all 22 module Makefiles and the CI workflow. gotestsum produces JUnit XML and JSON output natively, enabling proper test reporting in GitHub without extra conversion steps.                                                                                            
  - Add TEST_PACKAGES / COVER_PACKAGES variables to every Makefile so coverage only measures hand-written code, excluding generated code (ent ORM, oapi-codegen stubs, deepcopy, fakes, cmd entrypoints). This gives accurate coverage percentages per module.                                                      
  - Add -race flag consistently to all modules, exposing data races that were previously hidden.                                                                                                                                                                                                                    
  - Rework the CI test pipeline: replace gotestfmt-action, HTML/Cobertura report generation, and mikepenz/action-junit-report with a simpler flow — gotestsum for test execution, dorny/test-reporter for PR check annotations, per-module coverage summary in GitHub step summary, and a configurable coverage     
  threshold.                                                                                                                                                                                                                                                                                                        
  - Add coverage_threshold input (default: 70%) to the reusable workflow, with per-module overrides in ci.yaml for modules that are currently below the target.                                                                                                                                                     
  - Fix pre-existing bugs exposed by the new tooling:                                                                                                                                                                                                                                                               
    - Data race in notification/internal/sender/adapter/msteams — test helpers used unsynchronized counters across goroutines; fixed with sync/atomic.  
    - Data race in gateway/internal/controller/consumer_controller_test.go
    - Data race in identiy/internal/controller/suite_test.go


stefan-ctrl (approved):
LGTM

PR #349, opened 2026-04-21
feat: add GOTESTSUM_JUNIT_FLAGS for Makefiles to harmonize grouping i…

BjoernKarma:
## Why this change

`dorny/test-reporter` renders what it gets from `junit.xml`; it does not provide full layout templating. We saw inconsistent report grouping/count perception between runs (single-package vs multi-package), driven by JUnit naming/output shape rather than reporter-side layout settings.

## What changed

We standardized `gotestsum` JUnit options across module `Makefile` test targets by introducing a shared `GOTESTSUM_JUNIT_FLAGS` variable and reusing it in each `gotestsum` call.

Standardized flags:
- `--junitfile junit.xml`
- `--junitfile-testsuite-name relative`
- `--junitfile-testcase-classname relative`
- `--jsonfile gotest.log`

## Why `relative`

Using `relative` for suite/class names gives stable, readable identifiers across modules and runners while preserving package-level context. This reduces noise from full import paths and makes PR test reports more consistent when comparing single-package and multi-package runs.

### DIFF
--- a/admin/Makefile
+++ b/admin/Makefile
@@ -62,9 +62,11 @@ vet: ## Run go vet against code.
 TEST_PACKAGES = ./internal/...
 COVER_PACKAGES = ./internal/...
 
+GOTESTSUM_JUNIT_FLAGS = --junitfile junit.xml --junitfile-testsuite-name relative --junitfile-testcase-classname relative --jsonfile gotest.log
+
 .PHONY: test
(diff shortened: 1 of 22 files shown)


BjoernKarma:
This approach does not solve the problem. Closed in favor of: https://github.com/telekom/controlplane/pull/366

PR #350, opened 2026-04-22
feat(application): graceful client-secret rotation

ron96g:
Add the graceful client-secret rotation to all relevant domains:

- feat(admin): central configuration per zone for secret-rotation and notifications
- feat(common): reminder feature to send notifications based on some time triggers
- feat(common): added `ReconcileHint` to allow handlers to control the requeueAfter duration
- feat(application): core-logic that will perform the secret-rotation
  - locking of critical path to avoid collisions
  - handling of notifications for `secret-rotation-completed` and `secret-rotation-expiring`
  - handling of reminder-notification using new reminder feature in common
 - feat(secret-manager): added new secret-slot `rotatedClientSecret`
- fix(rover): never overwrite an application-secret after initial bootstrap
- docs: added docs for graceful secret rotation and notification attachments
- feat(install): updated local-overlay to deploy mailpit smtp-mock
- feat(hack): updated `local-setup.sh` to restart deployments that have been re-build


admin/api/v1/zone_types.go
```diff
@@ -31,9 +31,35 @@ type IdentityProviderAdminConfig struct {
...
 }
 
+type SecretRotationConfig struct {
+	// Enabled controls whether secret rotation is enabled for this zone.
+	//  If false, secrets will not be rotated and the grace period and expiration period will be ignored.
+	Enabled bool `json:"enabled"`
+	// GracePeriod is the duration that a rotated secret is valid for.
+	// This allows to have a smooth transition when rotating secrets, as the old secret is still valid for a certain period of time after rotation.
+	// +kubebuilder:validation:Required
+	GracePeriod metav1.Duration `json:"gracePeriod"`
```
lukas016, line 42: question: do we have validation for it somewhere? because i cannot find validation webhook for it.

ron96g: What did you have in mind with validation? I assumed that `metav1.Duration` already checks that its a valid duration string

lukas016: we have in current solution some min and max values. I think expiration period can be min 5 mins for rotated secret and max i don't remember.

And i don't see expiration time for new secret just for rotated. Is it correct because we have 2 expiration times now.

1. Expiration for new secret
2. Expiration for rotated (previous) secret

ron96g: There are 2 durations for expiration, however, that `RotationInterval` is not optimal. It infers that the CP already takes action - which it does not. The rotation still needs to be trigger externally. Renamed to `ExpirationPeriod`

application/internal/handler/application/handler.go
```diff
@@ -36,82 +43,211 @@ func (h *ApplicationHandler) CreateOrUpdate(ctx context.Context, app *applicatio
...
+	}
 
-		for _, failoverZone := range failoverZones {
-			err = CreateGatewayConsumer(ctx, failoverZone, app, WithFailover())
-			if err != nil {
-				return errors.Wrapf(err, "failed to create Gateway consumer for failover zone %s when creating application", failoverZone.Name)
-			}
+	for _, failoverZone := range failoverZones {
+		if _, err := CreateIdentityClient(ctx, failoverZone, app, WithFailover()); err != nil {
+			return nil, errors.Wrapf(err, "failed to create Identity client for failover zone %s when creating application", failoverZone.Name)
```
julius-malcovsky, line 134: should we use fmt instead of errors package?

application/internal/handler/application/notification.go
```diff
@@ -0,0 +1,175 @@
...
+	if app.Status.CurrentExpiresAt != nil {
+		properties["currentExpiresAt"] = app.Status.CurrentExpiresAt.Format(time.RFC3339)
+	}
+
+	notification, err := builder.New().
+		WithNamespace(app.Namespace).
+		WithOwner(app).
+		WithSender(notificationv1.SenderTypeSystem, "ApplicationService").
+		WithPurpose(PurposeRotationCompleted).
+		WithName(PurposeRotationCompleted).
```
julius-malcovsky: out of scope, but would it make sense to have the name less generic? so the purpose could be used for templates and then the name could have the name of the app for example. question is - useful or not worth it

common/pkg/controller/controller.go
```diff
@@ -106,18 +110,26 @@ func (c *ControllerImpl[T]) Reconcile(ctx context.Context, req reconcile.Request
...
+		c.Event(ctx, object, "Warning", "UnknownReady", "Resource has an unknown ready status")
 	}
 
 	StampObservedGeneration(object)
 	if err = c.Client.Status().Update(ctx, object); err != nil {
 		return HandleError(ctx, err, object, c.Recorder), nil
 	}
 
+	requeueAfter := config.RequeueWithJitter()
+	if hint.RequeueAfter != nil && *hint.RequeueAfter < requeueAfter {
```
julius-malcovsky, line 127: wondering if the *hint.RequeueAfter < requeueAfter makes sense. for normal cases yes, but for example with the approval expiry, it didnt make sense to check every x minutes (or even every 24 hours) - the timestamps wont change and if expiry is due in 7 months, thats a lot of wasteful cycles. 

for now its fine, but lets talk when we use this with approval expiry

julius-malcovsky (approved):
cosmetic remarks only, good work :) 
also discussed in person - feel free to merge :)

PR #358, opened 2026-04-23
fix: running golangci-lint run --fix across all modules

BjoernKarma:


BjoernKarma:
Will be done by module: Autofix and manual fix combined.

PR #366, opened 2026-04-27
fix: use Ginkgo native JUnit reports for accurate test counts in CI

BjoernKarma:
## Summary                                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                    
  - Use Ginkgo's `-ginkgo.junit-report` flag to produce per-spec JUnit XML for the 12 pure-Ginkgo modules, replacing gotestsum's `--junitfile` which only counts one test per package                                                                                                                               
  - Update `dorny/test-reporter` path to glob for `**/ginkgo-junit.xml`, falling back to `junit.xml` for modules that still use gotestsum's JUnit output                                                                                                                                                            
  - Keep gotestsum for terminal output (`--format pkgname`) and JSON logs (`--jsonfile gotest.log`) in all modules                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                    
  ## Why                                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                    
  All main modules use Ginkgo, which runs specs inside a single `testing.T` wrapper (e.g. `TestCondition`). From `go test -json`'s perspective there is only one test event per package — gotestsum's JUnit output inherits this limitation. For example, `controlplane-api` was reported as **6 tests** when it  actually has **110 Ginkgo specs**.
                                                                                                                                                                                                                                                                                                                    
  The `-ginkgo.junit-report` flag can only be used in modules where **all** test packages import Ginkgo, because non-Ginkgo test binaries reject the flag. 9 modules have a mix of both frameworks and remain on gotestsum's `--junitfile` for now. These could be migrated by either converting the remaining plain-Go test packages to Ginkgo suites or by splitting test runs per framework.
                                                                                                                                                                                                                                                                                                                    
  ## Modules affected                                                                                                                                                                                                                                                                                               
   
  | Change | Modules |                                                                                                                                                                                                                                                                                              
  |---|---|       
  | Ginkgo JUnit (per-spec counts) | admin, api, approval, common, controlplane-api, event, gateway, permission, projector, pubsub, rover, rover-ctl |                                                                                                                                                              
  | Unchanged (gotestsum, per-package counts) | application, common-server, discovery-server, file-manager, identity, notification, organization, rover-server, secret-manager, tools/snapshotter, tools/e2e-tester |


ron96g (approved):
LGTM. Lets align our test frameworks in the future so that we fully leverage this for coverage checks

PR #367, opened 2026-04-28
feat(go): use proper license file for go modules

lukas016:


.github/workflows/reusable-go-ci.yaml
```diff
@@ -256,6 +256,43 @@ jobs:
...
+            echo "${HELP_MESSAGE}"
+            ls -lh "${EXPECTED_SYMLINK_FILE}"
+            exit 1
+          fi
+
+          if [ "$(readlink -n "${EXPECTED_SYMLINK_FILE}")" != "${LICENSE_SOURCE_FILE}" ]; then
+            echo "${EXPECTED_SYMLINK_FILE} is missing or it isn't valid symlink to ${LICENSE_SOURCE_FILE}"
+            echo "${HELP_MESSAGE}"
+            exit 1
+          fi
```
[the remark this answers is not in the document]
lukas016: realpath or readlink -f doesn't work for me locally on wsl. It can be improved later, if someone complains.

admin/api/LICENSE.txt
```diff
@@ -0,0 +1 @@
+LICENSES/Apache-2.0.txt
```
BjoernKarma: What is the main goal we want to achieve with the license files? Do we need it for pkg.go.dev license detection?

I am asking, because it looks like the symlinks might not work on OS level due to the nested folder structure. The target is always `LICENSES/Apache-2.0.txt` regardless of the module and where it is in the nested folder structure. This might be okay if the lookup is done via git (e.g. github diff shows the content), but if the symlink is being followed on OS level, it will be dangling.

.github/workflows/reusable-go-ci.yaml
```diff
@@ -256,6 +256,43 @@ jobs:
...
+          declare -r HELP_MESSAGE="Please create symlink to license file: ln -s LICENSES/Apache-2.0.txt ${{ inputs.module }}/LICENSE.txt"
+          if [ ! -f go.mod ]; then
+            echo "WARNING: No go.mod found in ${{ inputs.name }}. Check license is skipped."
+            exit 0
+          fi
+
+          echo "Found go.mod. Checking ${EXPECTED_SYMLINK_FILE}"
+          if [ ! -L "${EXPECTED_SYMLINK_FILE}" ]; then
+            echo "${EXPECTED_SYMLINK_FILE} doesn't exist or it isn't symlink"
+            echo "${HELP_MESSAGE}"
```
BjoernKarma: What is the main goal we want to achieve with the license files? Do we need it for pkg.go.dev license detection?

I am asking, because CI check doesn't validate symlink resolution (e.g., with test -e "$EXPECTED_SYMLINK_FILE"). This means it will always pass as long as the symlink string matches, regardless of whether the license file is accessible.

.github/workflows/reusable-go-ci.yaml
```diff
@@ -256,6 +256,43 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: 1
+
+      - name: Check license file
+        shell: bash
+        working-directory: ${{ inputs.module }}
```
BjoernKarma: Is it intentional that we check the license file only for the main modules? For example nested API modules (like admin/api) are not covered by CI.

BjoernKarma (commented):
Looks good, I have added 3 questions/remarks. Maybe we can have a look together?

admin/api/LICENSE.txt
lukas016: correct it is for pkg.go.dev but i want have work symlink on os level too. I will fix it.

.github/workflows/reusable-go-ci.yaml
lukas016: this is on discussion but if we want to support all modules in pkg.go.dev we should cover submodules too.

lukas016:
We discussed PR on the call and It was simplified. Just symlink for common modules were created manually.

BjoernKarma (approved):
LGTM

PR #375, opened 2026-04-30
fix(permission): Fix golangci-lint issues

BjoernKarma:
## Update Makefile
 Update tool versions and golangci-lint target in Makefile.

 ## Autofixes for `permission/`
 Use make lint-fix to autofix issues

  ## Manual lint fixes for `permission/`

- Reorder imports across permission handler/controller tests and controller code to satisfy linting (incl. dot-import placement).
- Fix import-shadowing by renaming local variables previously named log to logger.
- Update permission/Makefile to install golangci-lint via the v2 module path and bump the pinned version; remove the module-local permission/.golangci.yml so the module uses the repo-root config.


julius-malcovsky (approved):
LGTM

PR #381, opened 2026-04-30
feat(rover-server,rover-ctl): correctly supported secret-rotation; added support for graceful-secret-rotation

ron96g:
This PR implements the Rover-CTL and Rover-Server support for the Graceful-Secret-Rotation (see https://github.com/telekom/controlplane/pull/350)

- feat(rover-ctl): support in reset-secret command
- feat(rover-server): support for graceful-secret-rotation handling
- fix(rover-server): always update secret via Application, not via Rover directly

Missing:
- add secret-info also to get-info resource
- fix breaking change in naming: `clientSecret` vs `currentSecretValue`


rover-ctl/pkg/config/config.go
```diff
@@ -42,4 +42,9 @@ func setDefaults() {
 	viper.SetDefault("token", "") // ROVER_TOKEN
 	viper.SetDefault(ConfigKeyTokenURL, "")
 	viper.SetDefault("access.token", "") // ROVER_ACCESS_TOKEN (used only for local testing)
+
+	// Polling defaults
+	viper.SetDefault("timeout.status", "30s")         // ROVER_TIMEOUT_STATUS
+	viper.SetDefault("timeout.secretRotation", "30s") // ROVER_TIMEOUT_SECRET_ROTATION
```
lukas016, line 48: question: cannot we do it more generic, what can be used more more subcommands? e.g.: wait or command-timeout, timeout?

rover-ctl/pkg/commands/get-secret/cmd.go
```diff
@@ -0,0 +1,79 @@
...
+
+	cmd.Cmd.RunE = cmd.Run
+
+	return cmd.Cmd
+}
+
+func (c *Command) Run(cmd *cobra.Command, args []string) error {
+	handler, err := handlers.GetHandler("Rover", "tcp.ei.telekom.de/v1")
+	if err != nil {
+		return errors.Wrap(err, "failed to get rover handler")
```
lukas016, line 54: didn't we discuss this errors.Wrap is deprected?  Change it at once won't be easy (with AI it can be easier) but noone will be capable do proper review for it. So if we wan to migrate to fmt.Errorf, question is: why should we do this problem worse?

rover-ctl/pkg/handlers/v0/secret.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+	defer resp.Body.Close()
+
+	err = common.CheckResponseCode(resp, http.StatusOK)
+	if err != nil {
+		return nil, err
+	}
+
+	var response SecretRotationStatusResponse
+	if err := json.NewDecoder(resp.Body).Decode(&response); err != nil {
```
lukas016, line 65: FYI: good job

rover-ctl/pkg/handlers/v0/secret.go
```diff
@@ -0,0 +1,113 @@
...
+	ctx, cancel := context.WithTimeout(ctx, viper.GetDuration("timeout.secretRotation"))
+	defer cancel()
+
+	ticker := time.NewTicker(viper.GetDuration("poll.interval"))
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil, errors.New("timed out waiting for secret rotation to converge")
```
lukas016, line 83: FYI: in newer version of golang performance of errors.New and fmt.Errorf  should be the same. So you can use everywhere fmt.Errorf()

rover-server/api/openapi.yaml
```diff
@@ -404,6 +414,42 @@ paths:
...
+        - Rover
+      summary: Get the secret rotation status for an Application
+      description: >
+        Returns the current secret rotation status for an Application,
+        including whether a graceful rotation is in progress, whether it has
+        converged, and the relevant expiry timestamps.
+      operationId: getRoverSecretRotationStatus
+      parameters:
+        - $ref: '#/components/parameters/RoverId'
+      responses:
```
lukas016, line 431: can we return here some extra code, which will identified if some rotation operation is still "running" (was requested)?

rover-ctl/pkg/config/config.go
ron96g: Not sure about generic here but I agree that we should add some more timeouts for general commands and request-timeouts

rover-ctl/pkg/commands/get-secret/cmd.go
ron96g: AI probably used it because its used everywhere else in the code. Maybe we can add this to a linter so that we are forced to migrate? (Atleast on new code)... Migration to the fmt.Errorf can be done later... Personally, I dont think its a huge different if we need to migrate 250 vs 300 errors.Wrap(...)

rover-ctl/pkg/handlers/v0/secret.go
ron96g: See comment above

rover-server/api/openapi.yaml
ron96g: Currently it returns 409, see https://github.com/telekom/controlplane/pull/381/changes#diff-0004da45dfdcca13127d16ccc6eb4d5c2694f450dcbd1d23362ca23908b4c08eR285 ... But I guess you mean make it visible in the Openapi, right? Then yes, that should be added as explicit error-case

lukas016: yes, i see extra code 409 in openapi for patch but not for get. Do we want to use just ProcessingState in response or identified over http status too some operation is still running?

rover-ctl/pkg/handlers/v0/secret.go
```diff
@@ -0,0 +1,113 @@
...
+func (s *SecretRotationStatusResponse) IsGone() bool {
+	return s.Gone
+}
+
+// GetSecretRotationStatus fetches the current secret rotation status for an application.
+func (h *RoverHandler) GetSecretRotationStatus(ctx context.Context, name string) (*SecretRotationStatusResponse, error) {
+	token := h.Setup(ctx)
+	url := h.GetRequestUrl(token.Group, token.Team, name, "secret", "status")
+
+	resp, err := h.SendRequest(ctx, nil, http.MethodGet, url)
```
lukas016: optional: is second argument body? it is recommended to use http.NoBody for better readability

rover-server/internal/controller/rover.go
```diff
@@ -249,43 +252,135 @@ func (r *RoverController) GetApplicationsInfo(ctx context.Context, params api.Ge
...
-		return res, problems.InternalServerError("Failed to generate new client secret", err.Error())
+		if problems.IsNotFound(err) {
+			return res, problems.NotFound(resourceId)
+		}
+	}
+
+	// Check if a rotation is already in progress for the current generation
+	rotationCond := meta.FindStatusCondition(app.Status.Conditions, applicationv1.SecretRotationConditionType)
+	rotationInProgress := rotationCond != nil && rotationCond.Reason == applicationv1.SecretRotationReasonInProgress
+	isStale := rotationCond != nil && rotationCond.ObservedGeneration < app.GetGeneration()
```
lukas016, line 285: edge case: cannot be this potential problem? I can request rotation of secret but idp operator will be killed in middle of operation. Someone else will update another application field (icto, email, ...), so app generation will be incremented and i can trigger rotation again.

lukas016: what is additional benefit of this extra protection?

rover-server/internal/controller/rover.go
```diff
@@ -249,43 +252,135 @@ func (r *RoverController) GetApplicationsInfo(ctx context.Context, params api.Ge
...
+			ProcessingState: api.ProcessingStateProcessing,
+			OverallStatus:   api.OverallStatusProcessing,
+		}, nil
+	}
+
+	rotationCond := meta.FindStatusCondition(app.Status.Conditions, applicationv1.SecretRotationConditionType)
+
+	processingState := api.ProcessingStateDone
+	overallStatus := api.OverallStatusComplete
+	if rotationCond != nil {
```
lukas016, line 354: optional: this can be simplified just on 1 if block:

```golang
	if rotationCond != nil && rotationCond.Reason != applicationv1.SecretRotationReasonSuccess {
     		processingState = api.ProcessingStatePending
			overallStatus = api.OverallStatusPending
   }
```

lukas016:
LGTM: main logic is good, just some minor comments stuffs which can be improved later.

rover-server/internal/controller/rover.go
ron96g: Theoretically yes, but then we lose the ProcessingStateProcessing state which is set if `rotationCond.Reason == applicationv1.SecretRotationReasonInProgress`

lukas016: sry i updated condition little but it should be cover because i am checking if reason isn't Success

BjoernKarma (dismissed):
LGTM - Only comment would be what copilot has already marked in `rover-server/internal/controller/rover.go` in three locations for the missing error return.

BjoernKarma (approved):

PR #386, opened 2026-05-04
feat(rover, application, admin): Add support for externalIds

iagarba:
Introduce scheme-tagged externalIds on Rover and Application CRs, with per-zone  validation policies on the Zone CR. Customer-facing rover files keep scalar fields (psiid, icto); rover-server translates them into internal externalIds on  write and back on read.

 What
  ----
  - admin: add ExternalIdPolicy{Scheme, Required, Pattern} to ZoneSpec.
  - rover/api, application/api: add ExternalId{Scheme, Id} and ExternalIds list
    to the respective specs, with listType=map + listMapKey=scheme for
    server-side merge.
  - rover handler: propagate Rover.spec.externalIds → Application.spec.externalIds
    verbatim inside the existing mutator closure.
  - rover webhook: validate externalIds against zone.spec.externalIdPolicies.
    Format is always checked; Required=true additionally enforces presence.
    Regex compilation cached via sync.Map. GetZone now records zone-not-found on
    the shared ValidationError instead of returning a free-standing BadRequest.
  - rover-server: add scalar psiid to the OpenAPI schema (Rover,
    RoverCreate/UpdateRequest, RoverResponse, ApplicationInfo). externalIds is
    NOT exposed on the REST surface. Translation helper lives in
    internal/mapper/externalids.go; fixes a pre-existing gap where icto was
    declared but never mapped.
  - docs: user-journey/applications, admin-journey/environments-and-zones,
    architecture/rover updated.


application/internal/webhook/v1/application_webhook.go
```diff
@@ -72,12 +79,48 @@ func (v *ApplicationCustomValidator) ValidateDelete(_ context.Context, _ *applic
...
+// reported as a validation failure.
+func (v *ApplicationCustomValidator) getZone(ctx context.Context, app *applicationv1.Application, env string) (*adminv1.Zone, error) {
+	if app.Spec.Zone.Name == "" {
+		return nil, nil
+	}
+
+	// Zones live in the environment namespace, matching the Rover webhook.
+	zoneRef := client.ObjectKey{
+		Name:      app.Spec.Zone.Name,
+		Namespace: env,
```
ron96g: app.Spec.Zone.Namespace

application/internal/webhook/v1/application_webhook.go
```diff
@@ -72,12 +79,48 @@ func (v *ApplicationCustomValidator) ValidateDelete(_ context.Context, _ *applic
...
+	}
+	zone := &adminv1.Zone{}
+	if err := v.client.Get(ctx, zoneRef, zone); err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil, apierrors.NewBadRequest(fmt.Sprintf("zone '%s' not found", app.Spec.Zone.Name))
```
ron96g: '%s' --> %q

ron96g: Why not use the `cerrors.NewValidationError` defined in the calling function?

application/internal/webhook/v1/externalids.go
```diff
@@ -0,0 +1,77 @@
...
+	"sync"
+
+	"k8s.io/apimachinery/pkg/util/validation/field"
+
+	adminv1 "github.com/telekom/controlplane/admin/api/v1"
+	applicationv1 "github.com/telekom/controlplane/application/api/v1"
+	cerrors "github.com/telekom/controlplane/common/pkg/errors"
+)
+
+var patternCache sync.Map
```
ron96g: Why global? We should avoid global vars if not required

ron96g: Also add a comment what maps to what

docs/docs/admin-journey/environments-and-zones.md
```diff
@@ -93,6 +93,49 @@ Zones can be configured with different visibility levels:
 - **World** — The zone is accessible from outside the platform (public-facing APIs).
 - **Enterprise** — The zone is accessible only within the organization's network.
 
+### External Identifier Policies
```
ron96g: I think this belongs more into the "Platform Governance Features"

docs/docs/architecture/rover.mdx
```diff
@@ -29,6 +29,17 @@ Users interact with the Rover domain through three paths:
 - **Identity domain** — Configures authentication settings.
 - **Approval domain** — Integrates approval requirements for exposures.
 
+## External Identifiers
```
ron96g: too detailed. Maybe remove completely

docs/docs/user-journey/applications.mdx
```diff
@@ -62,6 +62,32 @@ spec:
 
 Rover itself defines the primary `zone`. Failover is configured on individual API exposures/subscriptions (not as `spec.failoverZone` on Rover).
 
+## External Identifiers
+
+Rover files can carry business identifiers that tie an application back to upstream inventory systems. Today two scalar fields are supported:
```
ron96g: Maybe mark this as "due to downward compatibility" as in the future new IDs would be added via externalIDs

docs/docs/user-journey/applications.mdx
```diff
@@ -62,6 +62,32 @@ spec:
...
+spec:
+  zone: dataplane1
+  psiid: PSI-103596
+  icto: icto-12345
+```
+
+Whether an identifier is *required* and the exact regex applied to its value are configured **per zone** by the platform administrator. If a zone has no policy for a scheme, supplying the field is optional and unvalidated. When a zone enforces a scheme, a Rover without that field (or with a malformed value) is rejected at admission time.
+
+:::note
+`kubectl apply --dry-run=client` does not fire admission webhooks, so format errors will only surface on server-side apply.
```
ron96g: Not relevant for user-journey as the user will not use kubectl

docs/docs/user-journey/applications.mdx
```diff
@@ -62,6 +62,32 @@ spec:
...
+kind: Rover
+metadata:
+  name: checkout-service
+spec:
+  zone: dataplane1
+  psiid: PSI-103596
+  icto: icto-12345
+```
+
+Whether an identifier is *required* and the exact regex applied to its value are configured **per zone** by the platform administrator. If a zone has no policy for a scheme, supplying the field is optional and unvalidated. When a zone enforces a scheme, a Rover without that field (or with a malformed value) is rejected at admission time.
```
ron96g, line 85: " If a zone has no policy for a scheme, supplying the field is optional and unvalidated." is this a good idea? People could then set any weird value they want

ron96g: Maybe BLOCK if not configured for zone? Might be breaking... Or silently drop

rover-server/internal/mapper/externalids.go
```diff
@@ -0,0 +1,61 @@
...
+// new scalar-scheme pair is a one-line change.
+const (
+	PsiScheme  = "psi"
+	IctoScheme = "icto"
+)
+
+// RoverScalarsToExternalIds packs non-empty customer scalar fields into a
+// deterministic []ExternalId for the internal Rover CR. Returns nil when no
+// scalars are supplied.
+func RoverScalarsToExternalIds(psiid, icto string) []roverv1.ExternalId {
```
ron96g: Multiple params with the same type are confusing and might result in mix-ups, so prefer param-structs

rover-server/internal/mapper/externalids.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+	if icto != "" {
+		out = append(out, roverv1.ExternalId{Scheme: IctoScheme, Id: icto})
+	}
+	return out
+}
+
+// RoverExternalIdsToScalars projects a Rover's ExternalIds back onto the
+// customer-facing scalar fields. Unknown schemes are ignored.
+func RoverExternalIdsToScalars(ids []roverv1.ExternalId) (psiid, icto string) {
```
ron96g: Same as above. Maybe return a struct instead

rover-server/internal/mapper/externalids.go
```diff
@@ -0,0 +1,61 @@
...
+			icto = e.Id
+		}
+	}
+	return psiid, icto
+}
+
+// ApplicationExternalIdsToScalars is the Application-CR counterpart to
+// RoverExternalIdsToScalars. Kept separate because the Application API group
+// has its own ExternalId type.
+func ApplicationExternalIdsToScalars(ids []applicationv1.ExternalId) (psiid, icto string) {
```
ron96g: Dead code? Why is this needed?

rover/internal/webhook/v1/externalids.go
```diff
@@ -0,0 +1,91 @@
+// Copyright 2025 Deutsche Telekom IT GmbH
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package v1
```
ron96g, line 5: Same code as in application. Does it even make sense to check this in both webhooks or just validate it in Rover as we do for all others validations?

ron96g (changes requested):
Really nice draft. Especially the optimization for caching compiled patterns. Some minor thigns we should clarify

application/internal/webhook/v1/application_webhook.go
iagarba: Done!

application/internal/webhook/v1/application_webhook.go
iagarba: Done!

application/internal/webhook/v1/externalids.go
iagarba: patternCache holds compiled regexes keyed by pattern string. Entries are added on first use and reused across admission calls. Bounded in practice by the number of distinct patterns across all Zones' ExternalIdPolicies. For the caching to work, the variable needs to be on package level

docs/docs/architecture/rover.mdx
iagarba: removed!

docs/docs/user-journey/applications.mdx
iagarba: removed!

docs/docs/user-journey/applications.mdx
iagarba: I thought about it but then rejected it because it would mean that the externalIdsPolicies will always have to be configured as "allowList". Coming from roverctl, an unknown field would just be silently dropped but direct kubectl apply would work...

rover-server/internal/mapper/externalids.go
iagarba: Done!

rover-server/internal/mapper/externalids.go
iagarba: Done!

rover-server/internal/mapper/externalids.go
iagarba: not needed, dead and buried!

rover/internal/webhook/v1/externalids.go
iagarba: Only needed if Applications can be created directly, but we don't have that usecase, so it is redundant.

docs/docs/admin-journey/environments-and-zones.md
iagarba: removed!

application/internal/webhook/v1/application_webhook.go
```diff
@@ -59,7 +59,7 @@ func (d *ApplicationCustomDefaulter) Default(ctx context.Context, app *applicati
 
 	env, ok := controller.GetEnvironment(app)
 	if !ok {
-		return fmt.Errorf("application %s does not have an environment label", app.GetName())
+		return fmt.Errorf("application '%q' does not have an environment label", app.GetName())
```
ron96g: %q is used as replacement for `'%s'` so now I will look like this: `''my-value''` ;) Is basically is a "quoted-string"

docs/docs/user-journey/applications.mdx
ron96g: Thats true until we extends rover-ctl to support the new external-id structure, then it will be an issue :D We can postpone this until this becomes an actual feature

rover-server/internal/controller/suite_controller_test.go
```diff
@@ -46,6 +47,12 @@ var app *fiber.App
 var mockFileManager *filefake.MockFileManager
 var stores *store.Stores
 
+// Pin the local time zone to UTC so JSON snapshots that contain metav1.Time
+// values render consistently across developer machines and CI.
+func init() {
+	time.Local = time.UTC
```
ron96g: I think this is a bad idea as it will result in tests that do not reflect the actual code. Instead we must ensure that all our APIs return the UTC (rfc3933) timestamp. But can we also talk about it in the team

ron96g (changes requested):
Looks good. Just minor things. Also: Did you remove the admin-docs completely now regarding this? Should it be a "Platform Governance Feature"?

application/internal/webhook/v1/application_webhook.go
iagarba: Done!

rover-server/internal/controller/suite_controller_test.go
iagarba: Thank God you're so thorough with your reviews :)

iagarba: This is what Claude says about it:
 Good call — that init() masks a production-visible bug (users in non-UTC timezones would see
  differently-formatted timestamps from the API). Let me find where the response times are produced and fix
  them at source.

  Read 1 file (ctrl+o to expand)

● Four response-building sites. All pass time.Time values onward, and encoding/json for time.Time formats
  using its location. Normalize to UTC at every point where we construct the response:

● Update(rover-server/internal/mapper/status/response.go)

──────────────────────────────────────────────────────────────────────────────────────────────────────────────
 Edit file
 rover-server/internal/mapper/status/response.go

` // asUTC returns t in UTC so the API serializes timestamps in RFC 3339 with a
 // trailing "Z" regardless of the server's local time zone. metav1.Time values
 // round-tripped through JSON carry whatever zone was attached during
 // unmarshal; normalizing here keeps the wire format stable.
 func asUTC(t time.Time) time.Time {
	return t.UTC()
 }

 // MapResponse maps the status of a generic resource to a ResourceStatusResponse.
   func MapResponse(ctx context.Context, obj types.Object) (api.ResourceStatusResponse, error) {
     status := MapStatus(obj.GetConditions(), obj.GetGeneration())
...
     }
 
     return api.ResourceStatusResponse{
     CreatedAt:       obj.GetCreationTimestamp().Time,
     ProcessedAt:     processedAtTime,
     CreatedAt:       asUTC(obj.GetCreationTimestamp().Time),
     ProcessedAt:     asUTC(processedAtTime),
     State:           status.State,
     ProcessingState: status.ProcessingState,
     OverallStatus:   CalculateOverallStatus(status.State, status.ProcessingState),`
 
What do you think?

ron96g (approved):
Thanks for fixing the comments. Looks good to me

PR #387, opened 2026-05-04
feat(approval-expiration): approvals can now expire and notifications are sent before they do

julius-malcovsky:
Notification templates will be added in a separate request


approval/internal/handler/util/notification.go
```diff
@@ -226,6 +233,69 @@ func initializeProperties() map[string]any {
...
+	purposeStringBuilder.WriteString(DELIMITER)
+	purposeStringBuilder.WriteString(data.Action)
+	purposeStringBuilder.WriteString(DELIMITER)
+	purposeStringBuilder.WriteString("reminder")
+	purposeStringBuilder.WriteString(DELIMITER)
+	purposeStringBuilder.WriteString(string(data.Actor))
+	purpose := purposeStringBuilder.String()
+
+	// Build notification name: <purpose>--<targetName>--<hash>
+	nameStringBuilder := strings.Builder{}
```
lukas016, line 286: optional: if it is on hotpath, you can preallocated builder with Grow()

approval/internal/handler/util/notification.go
```diff
@@ -226,6 +233,69 @@ func initializeProperties() map[string]any {
...
+	for k, v := range requesterMap {
+		properties[strings.ToLower(k)] = v
+	}
+
+	deciderMap, err := extractDecider(data.Decider)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to extract decider data")
+	}
+	for k, v := range deciderMap {
+		properties[strings.ToLower(k)] = v
```
lukas016, line 270: can you have the same key in deciderMap and requesterMap? because deciderMap key will override requesterMap

approval/internal/handler/approvalexpiration/handler.go
```diff
@@ -0,0 +1,247 @@
...
+	var daysRemaining int64
+
+	expired := now.After(ae.Spec.Expiration.Time) || now.Equal(ae.Spec.Expiration.Time)
+	switch {
+	case expired:
+		reminderType = "expired"
+		daysRemaining = 0
+	case now.After(ae.Spec.DailyReminder.Time):
+		reminderType = "daily"
+		daysRemaining = int64(ae.Spec.Expiration.Time.Sub(now).Hours() / 24)
```
lukas016: optional: this looks as something what can be in "utils" package.

approval/api/v1/builder/builder.go
```diff
@@ -241,11 +241,13 @@ func (b *approvalBuilder) Build(ctx context.Context) (finalResult ApprovalResult
 	// Approval was found
 	if approvalExists {
 		log.V(2).Info("Approval exists")
-		isDenied := b.Approval.Spec.State == v1.ApprovalStateRejected || b.Approval.Spec.State == v1.ApprovalStateSuspended
+		isDenied := b.Approval.Spec.State == v1.ApprovalStateRejected ||
```
ron96g, line 244: I know that you explained it in the Review but is it really the goal to enforce this? Currently its just purely informational

approval/api/v1/approvalexpiration_types.go
```diff
@@ -0,0 +1,81 @@
...
+// ApprovalExpirationSpec defines the desired state of ApprovalExpiration
+type ApprovalExpirationSpec struct {
+	// Approval is a reference to the parent Approval resource
+	Approval types.ObjectRef `json:"approval"`
+
+	// Expiration is the absolute date when the approval expires
+	Expiration metav1.Time `json:"expiration"`
+
+	// WeeklyReminder is the date from which weekly reminders start
+	WeeklyReminder metav1.Time `json:"weeklyReminder"`
```
ron96g: Can we use the reminder logic in common for this? That was implemented for graceful-client-secret-rotation?

approval/internal/config/expiration.go
```diff
@@ -0,0 +1,54 @@
...
+	viper.AutomaticEnv()
+	viper.SetEnvPrefix("APPROVAL")
+
+	config := &ExpirationConfig{
+		ExpirationPeriodMonths:       viper.GetInt("EXPIRATION_PERIOD_MONTHS"),
+		LastMonthsWithWeeklyReminder: viper.GetInt("EXPIRATION_WEEKLY_REMINDER_MONTHS"),
+		LastWeeksWithDailyReminder:   viper.GetInt("EXPIRATION_DAILY_REMINDER_WEEKS"),
+	}
+
+	if config.ExpirationPeriodMonths <= 0 {
```
ron96g: This could be done using https://github.com/go-playground/validator, however, the bigger question here is, can we integrate this into the common-config logic? The story to align config-management is still in the backlog, but can we already think about this?

approval/internal/config/expiration.go
```diff
@@ -0,0 +1,54 @@
...
+
+package config
+
+import (
+	"github.com/pkg/errors"
+	"github.com/spf13/viper"
+)
+
+// ExpirationConfig holds the configuration for approval expiration
+type ExpirationConfig struct {
```
ron96g, line 18: Is this something we want to configure via deployment config or rather via Environment? So that these values are set in the Environment CR and we read it from there instead from some configmap?

approval/internal/controller/approvalexpiration_controller.go
```diff
@@ -0,0 +1,58 @@
...
+// +kubebuilder:rbac:groups=notification.cp.ei.telekom.de,resources=notificationchannels,verbs=get;list;watch
+
+// Reconcile is part of the main kubernetes reconciliation loop
+func (r *ApprovalExpirationReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	return r.Controller.Reconcile(ctx, req, &approvalv1.ApprovalExpiration{})
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *ApprovalExpirationReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	r.Recorder = mgr.GetEventRecorderFor("approvalexpiration-controller")
```
ron96g, line 47: This API is deprecated and we must migrate at some point. Maybe already use the new one for newer controllers? Dont forget about the kubebuilder:rbac annotation

approval/internal/handler/approval/handler.go
```diff
@@ -125,3 +146,121 @@ func (h *ApprovalHandler) Delete(ctx context.Context, approval *approvalv1.Appro
 	logger.Info("Approval deleted")
 	return nil
 }
+
+// filterSystemActions removes system-only actions (Expire) from the available transitions
+func filterSystemActions(transitions approvalv1.AvailableTransitions) approvalv1.AvailableTransitions {
+	filtered := make(approvalv1.AvailableTransitions, 0, len(transitions))
```
ron96g: This could allocate +1 slots if Expire is not added

approval/internal/condition/condition.go
```diff
@@ -54,3 +54,12 @@ func NewSemigrantedCondition() metav1.Condition {
 		Message: "Request has been partially approved, awaiting second approval",
 	}
 }
+
+func NewExpiredCondition() metav1.Condition {
+	return metav1.Condition{
+		Type:    "Approved",
+		Status:  metav1.ConditionTrue,
```
ron96g: Is this correct?

approval/internal/webhook/v1/approval_webhook.go
```diff
@@ -124,3 +128,19 @@ func (a *ApprovalCustomValidator) ValidateDelete(_ context.Context, obj *approva
...
+
+// validateExpireTransition blocks manual transitions to EXPIRED state (system-only action)
+func validateExpireTransition(newObj *approvalv1.Approval) error {
+	if newObj.Spec.State != approvalv1.ApprovalStateExpired {
+		return nil
+	}
+
+	// Check if this is a legitimate system transition
+	for _, decision := range newObj.Spec.Decisions {
+		if decision.Name == "System" && decision.ResultingState == approvalv1.ApprovalStateExpired {
```
ron96g: Should "System" be a const somewhere?

approval/internal/webhook/v1/approval_webhook.go
```diff
@@ -124,3 +128,19 @@ func (a *ApprovalCustomValidator) ValidateDelete(_ context.Context, obj *approva
...
+			return nil // Valid system transition
+		}
+	}
+
+	return apierrors.NewBadRequest("Expire action is system-only and cannot be triggered manually")
```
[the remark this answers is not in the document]
ron96g: Good point by AI ;)

ron96g (changes requested):
Looks solid. Just a few questions/open points.

Add some diagram that explains the logic in more details? Maybe inside of the README similiar to the other digrams there.

approval/api/v1/builder/builder.go
julius-malcovsky: not sure I understand the question :/ if we dont treat an expired one as denied wouldnt it result in access not being revoked ? maybe we can discuss it ...

approval/internal/handler/util/notification.go
julius-malcovsky: will take a look in more detail, might need to update a larger part of code so putting it on the known issues list

approval/internal/config/expiration.go
julius-malcovsky: we already use the pattern with a .cfg file and a configMapGenerator so I thought it would be good to be consistent. If you had something else in mind lets discuss it :)

approval/api/v1/builder/builder.go
ron96g: Yes, but that is how its currently works. Its a toothless tiger - purely informational. So maybe it could be a policy like 
```
enum ApprovalExpirationPolicy {
IGNORE
REJECT
}
```

or something like this? And the default would be IGNORE

approval/internal/config/expiration.go
ron96g: The less config-files, the better ;) But you are right, lets keep it simple for now and ideally just use good defaults so we dont event need these config-files... Later we can rethink how the config should be handled

approval/api/v1/common_types.go
```diff
@@ -46,7 +49,6 @@ const (
 	ApprovalStateGranted     ApprovalState = "Granted"
 	ApprovalStateRejected    ApprovalState = "Rejected"
 	ApprovalStateSuspended   ApprovalState = "Suspended"
-	ApprovalStateExpired     ApprovalState = "Expired"
```
ron96g, line 49: Why remove the State?

approval/internal/config/expiration.go
```diff
@@ -0,0 +1,81 @@
...
+	// ExpirationDuration is how long until an approval expires (from creation)
+	ExpirationDuration time.Duration
+
+	// DefaultThresholds defines the reminder schedule for approvals
+	DefaultThresholds []reminder.Threshold
+}
+
+// LoadExpirationConfig loads the expiration configuration from environment variables
+func LoadExpirationConfig() (*ExpirationConfig, error) {
```
ron96g, line 37: Will this clash with our current controller-config? I think we should either re-use the stuff there or create a new local Viper-Instance here, but mixing them could result in bugs

approval/internal/config/expiration.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	// Convert to durations
+	expirationDuration := time.Duration(expirationMonths) * 30 * 24 * time.Hour
+
+	// Build default thresholds
+	var thresholds []reminder.Threshold
+
+	// Weekly reminder threshold (if configured)
+	if weeklyReminderMonths > 0 {
```
ron96g: Why configure it like this and not via a object? I think Viper is capable of loading complex data like JSON-objects and lists... Maybe check this.

approval/internal/controller/approvalexpiration_controller.go
```diff
@@ -0,0 +1,83 @@
...
+		WithOptions(controller.Options{
+			MaxConcurrentReconciles: cconfig.MaxConcurrentReconciles,
+			RateLimiter:             cc.NewRateLimiter(),
+		}).
+		Complete(r)
+}
+
+// eventsRecorderAdapter adapts events.EventRecorder (new API) to record.EventRecorder (old API)
+// for compatibility with common controller until it's migrated to the new API.
+type eventsRecorderAdapter struct {
```
ron96g: Why is this here? I mean, I see the comment but for the time being we can just use both the old and new API until we can fully migrate common to the new one.

approval/internal/controller/approvalrequest_controller.go
```diff
@@ -2,7 +2,7 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-package controller //nolint:dupl // approval and approvalrequest controllers are intentionally similar
```
ron96g, line 5: Not sure about removing this but if the linter does not complain, fine by me

approval/internal/handler/approval/fsm.go
```diff
@@ -25,8 +25,8 @@ var simple = fsm.Transitions{
 
 var fourEyes = fsm.Transitions{
 	{Action: v1.ApprovalActionAllow, Src: []v1.ApprovalState{v1.ApprovalStatePending, v1.ApprovalStateRejected}, Dst: v1.ApprovalStateSemigranted},
-	{Action: v1.ApprovalActionDeny, Src: []v1.ApprovalState{v1.ApprovalStatePending, v1.ApprovalStateGranted, v1.ApprovalStateSemigranted}, Dst: v1.ApprovalStateRejected},
```
ron96g, line 28: This is just a line change....

Where are the transitions for Expiration?

approval/internal/handler/approval/handler.go
```diff
@@ -125,3 +150,98 @@ func (h *ApprovalHandler) Delete(ctx context.Context, approval *approvalv1.Appro
 	logger.Info("Approval deleted")
 	return nil
 }
+
+// handleExpiration manages the lifecycle of ApprovalExpiration resources
+func (h *ApprovalHandler) handleExpiration(ctx context.Context, approval *approvalv1.Approval, stateChanged bool) error {
```
ron96g, line 170: [nitpick] This could also be an interface that is then added via composition to the approval-handler

approval/internal/handler/approval/handler.go
```diff
@@ -125,3 +150,98 @@ func (h *ApprovalHandler) Delete(ctx context.Context, approval *approvalv1.Appro
...
+	// Only for non-Auto strategies
+	if approval.Spec.Strategy == approvalv1.ApprovalStrategyAuto {
```
ron96g: So AUTO-Approved never expire? Thats the intended logic which we also used in the past?

approval/internal/handler/approval/handler.go
```diff
@@ -125,3 +150,98 @@ func (h *ApprovalHandler) Delete(ctx context.Context, approval *approvalv1.Appro
...
+
+	err := c.Get(ctx, key, ae)
+	if err != nil {
+		// Doesn't exist, nothing to delete
+		return client.IgnoreNotFound(err)
+	}
+
+	// Exists, delete it
+	if err := c.Delete(ctx, ae); err != nil {
+		return errors.Wrap(err, "failed to delete ApprovalExpiration")
```
ron96g, line 244: Could also ignore NotFound here

approval/internal/handler/approvalexpiration/handler.go
```diff
@@ -0,0 +1,196 @@
...
+	"github.com/telekom/controlplane/approval/internal/condition"
+	"github.com/telekom/controlplane/approval/internal/handler/util"
+	"github.com/telekom/controlplane/common/pkg/reminder"
+	"github.com/telekom/controlplane/common/pkg/types"
+	"github.com/telekom/controlplane/common/pkg/util/contextutil"
+)
+
+// Handler handles ApprovalExpiration resources
+type Handler struct {
+	client client.Client
```
ron96g: Why client? Normally we use the one from the ctx

approval/internal/handler/approvalexpiration/handler.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+// CreateOrUpdate processes an ApprovalExpiration resource
+func (h *Handler) CreateOrUpdate(ctx context.Context, ae *v1.ApprovalExpiration) error {
+	logger := log.FromContext(ctx)
+
+	// Load parent Approval
+	approval, err := h.getParentApproval(ctx, ae)
+	if err != nil {
+		return errors.Wrap(err, "failed to get parent approval")
```
ron96g, line 39: IgnoreNotFound?

approval/internal/handler/approval/handler.go
```diff
@@ -125,3 +150,98 @@ func (h *ApprovalHandler) Delete(ctx context.Context, approval *approvalv1.Appro
...
+	expirationDate := now.Add(cfg.ExpirationDuration)
+
+	ae := &approvalv1.ApprovalExpiration{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("%s--expiration", approval.Name),
+			Namespace: approval.Namespace,
+		},
+	}
+
+	mutate := func() error {
```
ron96g, line 212: [minor] add some labels or copy from Approval CR

approval/internal/handler/approval/handler.go
```diff
@@ -125,3 +150,98 @@ func (h *ApprovalHandler) Delete(ctx context.Context, approval *approvalv1.Appro
...
+}
+
+// createOrUpdateApprovalExpiration creates or updates an ApprovalExpiration with fresh dates
+func createOrUpdateApprovalExpiration(ctx context.Context, c commonclient.JanitorClient, approval *approvalv1.Approval, cfg *config.ExpirationConfig) error {
+	now := time.Now()
+	expirationDate := now.Add(cfg.ExpirationDuration)
+
+	ae := &approvalv1.ApprovalExpiration{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("%s--expiration", approval.Name),
```
ron96g: [minor] suffix not really needed as CR kind already is called ApprovalExpiration

approval/internal/handler/util/notification.go
```diff
@@ -78,6 +81,10 @@ type NotificationData struct {
 	Scenario               NotificationScenario
 	Actor                  Actor
 	Action                 string
+	// Expiration-specific fields
+	ExpirationDate string
```
ron96g, line 89: Why not split it to keep it more clear what is present in which scenario

approval/internal/handler/util/notification.go
```diff
@@ -226,6 +233,69 @@ func initializeProperties() map[string]any {
...
+	// Example: approvalexpiration--subscribe--reminder--decider
+	purposeStringBuilder := strings.Builder{}
+	purposeStringBuilder.WriteString(strings.ToLower(data.Owner.GetObjectKind().GroupVersionKind().Kind))
+	purposeStringBuilder.WriteString(DELIMITER)
+	purposeStringBuilder.WriteString(data.Action)
+	purposeStringBuilder.WriteString(DELIMITER)
+	purposeStringBuilder.WriteString("reminder")
+	purposeStringBuilder.WriteString(DELIMITER)
+	purposeStringBuilder.WriteString(string(data.Actor))
+	purpose := purposeStringBuilder.String()
```
ron96g, line 283: Is there already a template for this?

approval/internal/handler/util/notification.go
```diff
@@ -226,6 +233,69 @@ func initializeProperties() map[string]any {
...
+
+	// Build notification name: <purpose>--<targetName>--<hash>
+	nameStringBuilder := strings.Builder{}
+	nameStringBuilder.WriteString(purpose)
+	nameStringBuilder.WriteString(DELIMITER)
+	nameStringBuilder.WriteString(data.Target.GetName())
+	name := nameStringBuilder.String()
+
+	notificationBuilder := builder.New().
```
ron96g, line 292: A bit of a controversial thing but can we actually send the same Notification to both teams? Like "hey teams, you share a subscription which is about to expire. As team-provider please re-grant" or something?

approval/internal/webhook/v1/approval_webhook.go
```diff
@@ -73,7 +73,7 @@ func (a *ApprovalCustomValidator) ValidateCreate(_ context.Context, obj *approva
 }
 
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
-func (a *ApprovalCustomValidator) ValidateUpdate(_ context.Context, oldObj, newObj *approvalv1.Approval) (warnings admission.Warnings, err error) {
+func (a *ApprovalCustomValidator) ValidateUpdate(ctx context.Context, oldObj, newObj *approvalv1.Approval) (warnings admission.Warnings, err error) {
```
ron96g, line 76: If linter does not care, I also do not care

approval/internal/handler/approval/handler.go
julius-malcovsky: I think it makes sense .. why would you re-approve something that was autoapproved. would mean that first sub is for free but after a year you need to do it manually.

approval/api/v1/common_types.go
julius-malcovsky: will not be used, because we would have to do extra handling and treat Expired as Granted, so the idea was to keep it Granted but we agreed to add a field into the status with a timestamp

approval/internal/config/expiration.go
julius-malcovsky: thanks, updated

approval/internal/config/expiration.go
julius-malcovsky: updated

approval/internal/handler/approval/fsm.go
julius-malcovsky: no longer using an expired state

approval/internal/handler/approvalexpiration/handler.go
julius-malcovsky: vibed too hard and missed some ai slop - sry, fixed

approval/internal/handler/util/notification.go
julius-malcovsky: technically yes, but it would be "breaking" the pattern we already have. we dont send the approval updates to both teams this way etc ... plus i think the builder would need a refactor, not sure if we can tell it to look for channels in multiple namespaces ... do you want to make an exception here?

approval/internal/handler/approval/handler.go
```diff
@@ -62,6 +81,16 @@ func (h *ApprovalHandler) CreateOrUpdate(ctx context.Context, approval *approval
...
 	}
 
+	// Handle ApprovalExpiration lifecycle (do this after conditions are set)
+	if err := h.handleExpiration(ctx, approval, stateChanged); err != nil {
+		return errors.Wrap(err, "failed to handle expiration")
+	}
+
+	// note the expiration of this approval in the status
+	approval.Status.ExpiresAt = &metav1.Time{
+		Time: time.Now().Add(h.expirationCfg.ExpirationDuration),
```
ron96g: Wont this result in the time being set to the latest reconcile? So it will keep updating and then this timestamp cannot match the actual expiration timestamp

approval/internal/handler/approval/handler.go
ron96g: The goal of the expiration is that people think about their subscribers and evaluate if its still needed. So I would say that also AUTO-approved approvals should be considered. But up for discussion. Lets talk about this with the team

approval/internal/handler/approvalexpiration/handler.go
```diff
@@ -0,0 +1,182 @@
...
+
+	approval, err := h.getParentApproval(ctx, ae)
+	if err != nil {
+		return errors.Wrap(err, "failed to get parent approval")
+	}
+	if approval == nil {
+		return nil
+	}
+
+	removed := meta.RemoveStatusCondition(&approval.Status.Conditions, "Expired")
```
ron96g: Purpose?

approval/internal/handler/approvalexpiration/handler.go
```diff
@@ -0,0 +1,182 @@
...
+// Returns nil, nil if the parent no longer exists.
+func (h *Handler) getParentApproval(ctx context.Context, ae *v1.ApprovalExpiration) (*v1.Approval, error) {
+	c := commonclient.ClientFromContextOrDie(ctx)
+	approval := &v1.Approval{}
+	key := client.ObjectKey{
+		Name:      ae.Spec.Approval.Name,
+		Namespace: ae.Spec.Approval.Namespace,
+	}
+	if err := c.Get(ctx, key, approval); err != nil {
+		return nil, client.IgnoreNotFound(err)
```
ron96g: client.IgnoreNotFound(err) should be handled on caller-side

approval/internal/handler/approvalexpiration/handler.go
```diff
@@ -0,0 +1,182 @@
...
+	daysRemaining := int64(ae.Spec.Expiration.Time.Sub(now).Hours() / 24)
+	if daysRemaining < 0 {
+		daysRemaining = 0
+	}
+
+	// Determine reminder type based on threshold
+	var reminderType string
+	switch {
+	case threshold.Repeat != nil:
+		reminderType = "daily" // Repeating thresholds are daily
```
ron96g: Is this valid for all cases? In theory you could send a notification which is repeated every week

approval/internal/handler/approvalexpiration/handler.go
```diff
@@ -0,0 +1,182 @@
...
+			Owner:     ae,
+			StateNew:  string(approval.Spec.State),
+			StateOld:  string(approval.Spec.State), // State hasn't changed yet for reminders
+			Target:    &approval.Spec.Target,
+			Requester: &approval.Spec.Requester,
+			Decider:   &approval.Spec.Decider,
+			Scenario:  util.NotificationScenarioUpdated,
+			Action:    approval.Spec.Action,
+		},
+		ExpirationDate: ae.Spec.Expiration.Format("2006-01-02"),
```
ron96g: We should align timestamps to avoid confusion. I would suggest using an RFC 3339 "2026-06-09T08:13:00Z"

approval/internal/handler/util/notification.go
ron96g: Alright, lets do it this way. Should work fine. Maybe this could be an optimization in the future to reduce the amount of notifications that are sent

ron96g (commented):
Lets have a call to talk about this. I also have some suggestions regarding the viper-config.

approval/internal/handler/approval/handler.go
julius-malcovsky: you are right! 
time is now only set when changing into GRANTED state from a non-SUSPENDED state

approval/internal/handler/approval/handler.go
julius-malcovsky: when put like that i think it makes sense. since the current idea is just to send notifications, it would be a nice reminder for people to maintain their approvals, not forget about it - will remove the exception

approval/internal/handler/approvalexpiration/handler.go
julius-malcovsky: dead code, forgot to remove

PR #388, opened 2026-05-08
feat(identity): add managed client scopes with custom claim injection

ron96g:
Introduce ConfigureClientScopes to declaratively manage a Keycloak
client scope per realm, supporting two claim types:

- HardcodedClaim: injects a static value into tokens
- SessionNote: maps a Keycloak session note (e.g. clientId) into tokens

The reconciler creates/updates/deletes protocol mappers to match the
desired ClaimConfig list on the Realm CR, and assigns the scope as a
realm default so all clients inherit the claims automatically.

Includes ClaimConfig/ClaimType API types, protocol mapper builders,
full unit tests, and updated mocks.


BjoernKarma (approved):

PR #394, opened 2026-05-12
fix(application,identity): handle secret rotation for non-graceful identity clients

ron96g:
Complete rotation immediately when the identity client has graceful
secret-rotation disabled (disable-secret-rotation annotation), instead
of waiting indefinitely for expiry timestamps that will never arrive.
Also:
- Read SecretExpiresAt directly from Keycloak's expiration time attribute
  instead of computing it from creation time + expiration period
- Detect rotation requests via admission operation type instead of
  generation number, fixing edge cases on first update
- Propagate RotatedExpiresAt and RotatedClientSecret from identity
  client status to application status
- Extract resolveRotationOpts helper in the webhook mutator

### DIFF
--- a/application/internal/handler/application/handler.go
+++ b/application/internal/handler/application/handler.go
@@ -84,7 +84,7 @@ func (h *ApplicationHandler) CreateOrUpdate(ctx context.Context, app *applicatio
 	}
 
 	if rotationInProgress {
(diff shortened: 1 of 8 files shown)


BjoernKarma (approved):

PR #400, opened 2026-05-12
feat(admin,organization): add realm claims, rover realm, and refactor TeamApis to ManagedRoutes

ron96g:
- Add originZone, originStargate (HardcodedClaim) and clientId
  (SessionNote) claims to the default identity realm so all tokens
  carry zone-of-origin metadata.
- Create a dedicated "rover" identity realm per zone for internal
  admin-config clients (InternalIdentityRealm in ZoneStatus).
- Replace ApiConfig/TeamApiConfig with ManagedRouteConfig/ManagedRoutesConfig,
  introducing a required ManagedRouteType field (TeamAPI or Proxy).
  TeamAPI routes behave as before (auth, no ACL, team-api realm).
  Proxy routes are pure passthrough on the default gateway realm.
- Rename ZoneSpec.TeamApis to ManagedRoutes and ZoneStatus.TeamApiRoutes
  to ManagedRoutes across admin and organization modules.
- Update organization field index to filter on spec.managedRoutes and
  only match zones with at least one TeamAPI-type route.
- Remove emoji from remoteorganization error messages.


gateway/internal/handler/realm/handler.go
```diff
@@ -40,26 +52,41 @@ func (h *RealmHandler) Delete(ctx context.Context, realm *gatewayv1.Realm) error
...
-		return errors.Wrapf(err, "failed to create route '%s'", RouteTypeIssuer)
+		return errors.Wrapf(err, "failed to create route %q", gatewayv1.RouteTypeIssuer)
+	}
+	if route != nil {
+		realm.Status.IssuerRoute = types.ObjectRefFromObject(route)
+		realm.Status.IssuerUrl = route.Spec.Downstreams[0].Url()
+	} else {
+		realm.Status.IssuerRoute = nil
+		realm.Status.IssuerUrl = ""
 	}
```
stefan-ctrl, line 70: I find the double `nil `check (first `err`, then the `route`) awkward.

Should we consider writing it in a different way, as this approach looks like some kind of code smell?
It encodes a business decision (don't create a route), into something that can be easily forgotten to check.

Normally, I would expect, if `err == nil`, then the _targeted_ return value (in this case `route`) would be there, and I would not need to check it again.


I'd prefer something more direct and checking with `errors.Is` or `errors.As`:

```go
// route.go
var ErrRouteDisabled = errors.New("route is disabled")

// handler.go
if errors.Is(err, ErrRouteDisabled) {
    realm.Status.CertsRoute = nil
    realm.Status.CertsUrl = ""
} else if err != nil {
  // ...
} else {
  // ...
}
```

stefan-ctrl (commented):
Great job. 
Only have a minor comment regarding nil check in route-`handler.go`.

gateway/internal/handler/realm/handler.go
ron96g: Nice improvement. Implemented

stefan-ctrl (approved):
LGTM!

PR #404, opened 2026-05-25
Event exposures need to watch applications

julius-malcovsky:
It can happen that a new event exposure gets stuck waiting for an application to become ready. The application actually is ready, but the exposure doesnt watch the CR.


stefan-ctrl (approved):
LGTM

PR #419, opened 2026-06-08
fix(secret-manager): invalidate cached children on parent update

ron96g:
This PR adds the fix in the secret-manager cached backend to invalidate all **known** children on parent change. A child in known if it has been requested via Get or Set previously. 

This implementation assumes the amount of children is small as it will allocate a new entry in a sync.Map for each.

### DIFF
--- a/secret-manager/pkg/backend/cache/backend.go
+++ b/secret-manager/pkg/backend/cache/backend.go
@@ -115,6 +115,20 @@ func (c *CachedBackend[T, S]) registerChild(id T) {
 	actual.(*sync.Map).Store(childKey, struct{}{})
 }
 
+// unregisterChild removes a sub-secret's cache key from its parent's tracking
+// set. This prevents unbounded growth of the children map when sub-secrets are
+// deleted without the parent being updated.
+func (c *CachedBackend[T, S]) unregisterChild(id T) {
+	if id.SubPath() == backend.NoSubPath {
+		return
+	}
+	parentKey := id.ParentId().CacheKey()
+	childKey := id.CacheKey()
+	if childMap, ok := c.children.Load(parentKey); ok {
+		childMap.(*sync.Map).Delete(childKey)
+	}
+}
+
 // invalidateChildren removes all cached sub-secret entries derived from a parent
 // secret when the parent is written or deleted. This prevents stale reads of
 // sub-secrets after the parent document is updated directly.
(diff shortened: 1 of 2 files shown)


secret-manager/pkg/backend/cache/backend_test.go
```diff
@@ -466,17 +469,128 @@ var _ = Describe("Cached Backend cache", func() {
...
+			parentSecretId := mocks.NewMockSecretId(GinkgoT())
+			parentSecretId.EXPECT().CacheKey().Return("env:team:app:externalSecrets")
+			parentSecretId.EXPECT().String().Return("env:team:app:externalSecrets:").Maybe()
+			parentSecretId.EXPECT().Copy().Return(parentSecretId).Once()
+			parentSecretId.EXPECT().SubPath().Return("")
+
+			// Create two sub-secret IDs
+			subSecretId1 := mocks.NewMockSecretId(GinkgoT())
+			subSecretId1.EXPECT().CacheKey().Return("env:team:app:externalSecrets/key1")
+			subSecretId1.EXPECT().String().Return("env:team:app:externalSecrets/key1:abc").Maybe()
```
stefan-ctrl: I'm wondering:

Why does the parentSecretID ends with a `:` in `parentSecretId.EXPECT().String().Return("env:team:app:externalSecrets:").Maybe()`, but later
is checked against `/ in `env:team:app:externalSecrets/key1`?

Additionally: Why are these calls only `Maybe()` expected?

secret-manager/pkg/backend/cache/backend.go
```diff
@@ -199,7 +235,9 @@ func (c *CachedBackend[T, S]) Set(ctx context.Context, id T, value backend.Secre
 		log.Info("Failed to add item to cache", "id", cacheKey)
 	}
 	c.group.Forget(cacheKey)
+	c.registerChild(id)
```
stefan-ctrl: Minor:
The order seems kinda wrong: to first register and then invalidate.

Might want to flip it and move the invalidation to the top of the function?

secret-manager/pkg/backend/cache/backend_test.go
ron96g: These are different calls (Maybe for logging). Secret-IDs are allowed to end with ":" when the checksum is unknown

secret-manager/pkg/backend/cache/backend.go
ron96g: They are mutually exclusive, so one is only executed for Secrets with SubSecrets and the other one for Secrets without SubSecrets. I still flipped it because it did read a bit weird :D

stefan-ctrl (approved):
Feedback implemented. And otherwise looks very good. 👍

PR #421, opened 2026-06-10
feat(admin): improve the bootstrapping of a zone by restructuring the code

ron96g:


julius-malcovsky (approved):
Cosmetic idea - maybe mark the functions that are used as a Step (a comment?) so that its easy to find them and they stand out a bit .. for future development/fixes

PR #422, opened 2026-06-10
refactor(admin, api, application, route-tester): decouple realmName from Environment

stefan-ctrl:
- refactor(admin, api, application, route-tester): decouple realmName
- refactor(admin): remove unused foo field
- test(admin): add a test to check specifically for decoupling


admin/config/certmanager/issuer.yaml
```diff
@@ -0,0 +1,17 @@
+# Copyright 2026 Deutsche Telekom IT GmbH
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# The following manifest contains a self-signed issuer CR.
+# More information can be found at https://docs.cert-manager.io
+# WARNING: Targets CertManager v1.0. Check https://cert-manager.io/docs/installation/upgrading/ for breaking changes.
+apiVersion: cert-manager.io/v1
```
ron96g: Remove this as we always use the central controlplane-issuer

admin/api/v1/environment_types.go
```diff
@@ -13,7 +13,15 @@ import (
...
-	Foo string `json:"foo,omitempty"`
+
+	// RealmName overrides the default realm name (which is the environment name).
+	// When set, identity and gateway realms will be created with this name instead
+	// of deriving it from the environment's metadata.name.
+	// This field is immutable after first reconciliation.
+	// +kubebuilder:validation:Pattern=`^[a-z0-9]+(-[a-z0-9]+)*$`
+	// +kubebuilder:validation:MaxLength=63
+	// +required
+	RealmName string `json:"realmName"`
```
ron96g: Why not default to Environment name?

api/internal/handler/apiexposure/handler.go
```diff
@@ -94,6 +93,14 @@ func (h *ApiExposureHandler) CreateOrUpdate(ctx context.Context, apiExp *apiapi.
 		return errors.Wrapf(err, "failed to find cross-zone subscription zones for apiExposure: %s", apiExp.Name)
 	}
 
+	// Resolve realm name from the zone status
+	scopedClient := cclient.ClientFromContextOrDie(ctx)
+	exposureZone, err := util.GetZone(ctx, scopedClient, apiExp.Spec.Zone.K8s())
+	if err != nil {
+		return errors.Wrapf(err, "failed to get zone %s for apiExposure: %s", apiExp.Spec.Zone.Name, apiExp.Name)
```
ron96g: Does this already handle NotFound as Blocked via util.GetZone(...)?

api/internal/handler/apisubscription/handler.go
```diff
@@ -214,6 +213,13 @@ func (h *ApiSubscriptionHandler) CreateOrUpdate(ctx context.Context, apiSub *api
 	// - Provider failover zone: use the failover route from ApiExposure.Status.FailoverRoute
 	// - Cross-zone: find matching proxy route in ApiExposure.Status.ProxyRoutes
 
+	// Resolve realm name from the exposure zone status
+	exposureZone, err := util.GetZone(ctx, scopedClient, apiExposure.Spec.Zone.K8s())
+	if err != nil {
+		return errors.Wrapf(err, "failed to get exposure zone %s", apiExposure.Spec.Zone.Name)
```
ron96g: Same as above. Blocked error-handling

api/internal/handler/util/getters.go
```diff
@@ -121,6 +121,15 @@ func GetRealmForZone(ctx context.Context, zoneRef types.ObjectRef, realmName str
 	return realm, zone, nil
 }
 
+// RealmNameForZone returns the gateway realm name from the Zone status.
+// This decouples the realm name from the environment name.
+func RealmNameForZone(zone *adminapi.Zone) string {
```
ron96g: Not sure about this method? Empty string feels wrong and also the whole function feels unnecessary

ron96g (changes requested):
Looks good. Just some minor topics. Just not 100% sure about calling it RealmName. I mean its correct, but it feels really specific. But something like "InternalName" might be too generic

admin/config/certmanager/issuer.yaml
stefan-ctrl: Thanks. Removed.

admin/api/v1/environment_types.go
stefan-ctrl: I thought about this as well, but I liked the style of explicit configuration more when I implemented it.

Thinking about this now, falling back to `metadata.name` makes more sense, since virtual environments would then have a lot of duplicated information.

Implemented `metadata.name` as a default fallback.

api/internal/handler/util/getters.go
stefan-ctrl: Thanks for the hint. Removed it and call `exposureZone.Status.GatewayRealm.Name` directly

admin/config/certmanager/kustomization.yaml
```diff
@@ -0,0 +1,10 @@
+# Copyright 2026 Deutsche Telekom IT GmbH
+#
+# SPDX-License-Identifier: Apache-2.0
+
+resources:
+- certificate-webhook.yaml
+#- certificate-metrics.yaml
```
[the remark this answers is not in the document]
stefan-ctrl: The issuer has been removed. See other comments.

admin/config/certmanager/issuer.yaml
```diff
@@ -0,0 +1,17 @@
...
+kind: Issuer
+metadata:
+  labels:
+    app.kubernetes.io/name: admin
+    app.kubernetes.io/managed-by: kustomize
+  name: selfsigned-issuer
+  namespace: system
+spec:
+  selfSigned: {}
```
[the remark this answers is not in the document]
stefan-ctrl: The issuer has been removed. See other comments.

api/internal/handler/util/getters.go
```diff
@@ -121,6 +121,15 @@ func GetRealmForZone(ctx context.Context, zoneRef types.ObjectRef, realmName str
...
+	if zone.Status.GatewayRealm != nil {
+		return zone.Status.GatewayRealm.Name
+	}
+	return ""
+}
```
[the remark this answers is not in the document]
stefan-ctrl: function has been removed, as deemed unnecessary.

admin/internal/handler/util/naming/naming.go
```diff
@@ -16,23 +16,23 @@ const (
 )
 
 func ForDefaultIdentityRealm(environment *adminv1.Environment) string {
-	return environment.GetName()
+	return environment.Spec.RealmName
 }
```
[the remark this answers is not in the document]
stefan-ctrl: Changed. See https://github.com/telekom/controlplane/pull/422/changes/043179a4b17316e796251e16be2a63b6a77168c6#diff-4a7e833abb82f6d2187d705e55bc8aafc38e87162b78b872586478826fc822b4

tools/route-tester/main.go
```diff
@@ -82,12 +83,24 @@ func main() {
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      zoneName,
+			Namespace: environment,
+		},
+	}
+	err = k8sClient.Get(ctx, client.ObjectKeyFromObject(zone), zone)
+	if err != nil {
+		panic(errors.Wrapf(err, "failed to get Zone %s in namespace %s", zoneName, environment))
+	}
+	realmName := zone.Status.GatewayRealm.Name
```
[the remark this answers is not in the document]
stefan-ctrl: Implemented in slight variation

application/internal/handler/application/handler.go
```diff
@@ -280,12 +280,11 @@ func CreateIdentityClient(ctx context.Context, zone *admin.Zone, owner *applicat
 	c := client.ClientFromContextOrDie(ctx)
 	clientId := MakeClientName(owner)
 	resourceName := clientId + "--" + zone.Name
-	realmName := contextutil.EnvFromContextOrDie(ctx)
 
-	// get namespace from zoneStatus
+	// Resolve realm name from zone status (decoupled from environment name)
+	realmName := zone.Status.IdentityRealm.Name
 	namespace := zone.Status.Namespace
```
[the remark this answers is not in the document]
stefan-ctrl: Implemented, but with a simpler check

api/internal/handler/apiexposure/handler.go
stefan-ctrl: Yes: 

https://github.com/telekom/controlplane/blob/5e666cf4bd6d63ccbfb55a4a48c117844f2a3d74/api/internal/handler/util/getters.go#L37-L49

But to be sure, since I'm not that familiar with error stack handling: `errors.Wrapf` is enough to "escalate" the blocked error, right?

ron96g: Yes, Wrapping and fmt.Errorf("some message: %w", err) will both propegate the cause

julius-malcovsky (dismissed):
LGTM

ron96g (changes requested):
There are also routes being created in event. These probably also need an update. They use `adminv1.RealmNameFromContext(ctx)`. We should then remove this function as it should no longer be used... After that is done and everything looks good, lets merge this so Julius can directly use it for agentic.

ron96g (approved):

PR #430, opened 2026-06-17
feat(agentic): Add support for AI related exposures and subscriptions (mcp + telecontext)

julius-malcovsky:


admin/api/v1/zone_types.go
```diff
@@ -94,6 +94,18 @@ const (
 	ManagedRouteTypeProxy ManagedRouteType = "Proxy"
 )
 
+// AiGatewayConfig configures an optional AI Gateway for this zone.
+// When present, the zone supports MCP (Model Context Protocol) exposures
+// that are routed through a dedicated AI Gateway instance with streaming support.
+type AiGatewayConfig struct {
```
ron96g, line 276: Controversial Question: Is this a core feature? Should it be configured via the Zone or some `AgenticZoneConfig` CR?

agentic/api/v1/mcpexposure_types.go
```diff
@@ -0,0 +1,131 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// McpExposureSpec defines the desired state of McpExposure.
+type McpExposureSpec struct {
+	// McpBasePath references the McpServer via its basePath.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:Pattern=`^/[a-z0-9-/]+$`
+	McpBasePath string `json:"mcpBasePath"`
```
ron96g: Why not jsut basePath? MCP is already clear from the context

agentic/api/v1/mcpexposure_types.go
```diff
@@ -0,0 +1,131 @@
...
+	// Approval configures how subscriptions to this MCP server are approved.
+	Approval Approval `json:"approval"`
+
+	// Zone references the Zone CR where this MCP server is exposed.
+	Zone ctypes.ObjectRef `json:"zone"`
+
+	// Provider identifies the providing application.
+	Provider ctypes.ObjectRef `json:"provider"`
+
+	// Variant defines the exposure variant (MCP or TELECONTEXTMCP).
```
ron96g, line 39: `TELECONTEXTMCP` is a product name, not a super big fan. We need a better name

agentic/api/v1/mcpexposure_types.go
```diff
@@ -0,0 +1,131 @@
...
+	Route *ctypes.ObjectRef `json:"route,omitempty"`
+
+	// ProxyRoutes references proxy gateway Route CRs for cross-zone MCP delivery.
+	// +optional
+	ProxyRoutes []ctypes.ObjectRef `json:"proxyRoutes,omitempty"`
+
+	// TelecontextConsumeRoute references the auto-created ConsumeRoute for Telecontext.
+	// Only set when variant is TELECONTEXTMCP.
+	// +optional
+	TelecontextConsumeRoute *ctypes.ObjectRef `json:"telecontextConsumeRoute,omitempty"`
```
ron96g: Why not add this as a `spec.security.defaultConsumers`? That way it would gain access without creating an extra CR

agentic/api/v1/mcpexposure_types.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="BasePath",type="string",JSONPath=".spec.mcpBasePath",description="The MCP server base path"
+// +kubebuilder:printcolumn:name="Active",type="boolean",JSONPath=".status.active",description="Whether this exposure is active"
+// +kubebuilder:printcolumn:name="Variant",type="string",JSONPath=".spec.variant",description="The exposure variant"
+// +kubebuilder:printcolumn:name="CreatedAt",type="date",JSONPath=".metadata.creationTimestamp",description="Creation timestamp"
```
ron96g: I think the default name of this column is Age

agentic/api/v1/security_types.go
```diff
@@ -0,0 +1,118 @@
+// Copyright 2026 Deutsche Telekom IT GmbH
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package v1
```
ron96g, line 5: Maybe really something we could put into common but different topic

common-server/LICENSE.txt
ron96g, line 1: what changed? Same for the other file

common/pkg/config/feature.go
```diff
@@ -48,6 +48,7 @@ var (
 	FeaturePermission    Feature = NewFeature("permission", false)    // Permission feature disabled by default
 	FeatureSecretManager Feature = NewFeature("secret_manager", true) // Secret Manager feature enabled by default
 	FeatureFileManager   Feature = NewFeature("file_manager", true)   // File Manager feature enabled by default
+	FeatureAiGateway     Feature = NewFeature("ai_gateway", false)    // AI Gateway (MCP) feature disabled by default
```
ron96g, line 51: Same question as in admin. Should this be a core-feature or like event "addon-feature"?

rover/internal/handler/rover/ai/exposure.go
```diff
@@ -0,0 +1,296 @@
+// Copyright 2026 Deutsche Telekom IT GmbH
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package ai
```
ron96g, line 5: not agentic?

rover/api/v1/rover_types.go
```diff
@@ -242,19 +251,25 @@ func (e *Exposure) Type() Type {
...
+	if e.Ai != nil {
+		return TypeAi
+	}
 	return ""
 }
 
 // Subscription defines a service that this Rover consumes
-// +kubebuilder:validation:XValidation:rule="self == null || has(self.api) || has(self.event)", message="At least one of api or event must be specified"
-// +kubebuilder:validation:XValidation:rule="(has(self.api) && !has(self.event)) || (!has(self.api) && has(self.event))", message="Only one of api or event can be specified (XOR relationship)"
+// +kubebuilder:validation:MaxProperties=1
```
ron96g, line 291: And this really works as expected?

ron96g (commented):
First review - more will follow.

I think what really is obvious is the amount of duplicate code. Its not only in the API but also in rover and rover-server. Does not feel so nice. Its only boilerplate but still I would say more than 50% of the lines of code are duplicate :-/ 

However, some I also dont fully understand, e.g. why is mapAiRateLimitand  others needed? Isnt the underlying structure for RateLimit shared for both API and Agentic?

Also the naming seems inconsistent, why is it sometimes called AI and then other times Agentic? Is there a reason or just mistake?

admin/api/v1/zone_types.go
julius-malcovsky: Good question - and a valid point. To be brutally honest, right now the AIGateway is rather simple and only creates a Gateway + a realm .. well the realm will be gone very soon and then only the gateway remains, which feels like an overkill to have a dedicated CR + controller + handler for that. BUT! If we want to look at our features as manadatory/optional (core/.. no core?) .. then lets not do size comparison and let it not matter if its 3 or 300 lines of code that handles them. It will be much easier to remember: optional? go look for a dedicated CR that configures this feature. Core? look into admin/zone. Does this align with what you had in mind ?

agentic/api/v1/mcpexposure_types.go
julius-malcovsky: all right :)

agentic/api/v1/mcpexposure_types.go
julius-malcovsky: AutoSubscribed? PlatformMCP? TeleMCP? 

Also, bear in mind that if we change it we either make it a breaking change or it will not be backwards compatible :(

agentic/api/v1/mcpexposure_types.go
julius-malcovsky: good idea :)

agentic/api/v1/mcpexposure_types.go
julius-malcovsky: thanks, fixed

common-server/LICENSE.txt
julius-malcovsky: seems to be an issue with my local checkout of the repo, perhaps my wsl volume is broken - shouldnt be here - ignore it pls

common/pkg/config/feature.go
julius-malcovsky: lets discuss it in the first comment, if we move it to a dedicated CR this will be gone too

rover/api/v1/rover_types.go
julius-malcovsky: there was a problem with the cost of the validation expression, therefore its removed and replaced by this .. will test it / fix it as needed

rover/internal/handler/rover/ai/exposure.go
julius-malcovsky: its a bit complicated .. there are many references to this that use the name AI 

What changes:

 - package ai → package agentic (directory + package name)
 - TypeAi → TypeAgentic, "ai" → "agentic" (type string value)
 - AiExposure → AgenticExposure, AiSubscription → AgenticSubscription
 - Exposure.Ai → Exposure.Agentic (field + JSON tag)
 - RoverStatus.AiExposures/AiSubscriptions → AgenticExposures/AgenticSubscription

do we want to change all of them? I believe here we could make an expcetion and break backwards compatibility, but do we want an exp/sub type agentic or should we stick to ai ? 

your original question is ok, yes, that one could be agentic .. but what about the others?

admin/internal/handler/zone/handler.go
```diff
@@ -65,6 +73,17 @@ func (h *ZoneHandler) CreateOrUpdate(ctx context.Context, obj *adminv1.Zone) err
...
 	}
 
+	// AI Gateway configuration
+	if cconfig.FeatureAiGateway.IsEnabled() && obj.Spec.AiGateway != nil {
+		if err := reconcileAiGateway(ctx, hc, obj); err != nil {
+			return err
+		}
+	} else {
+		obj.Status.AiGateway = nil
+		obj.Status.AiGatewayRealm = nil
```
ron96g: Remove all ref to GatewayRealm as no longer exists

common-server/LICENSE.txt
ron96g: should be deleted. Same as below

ron96g (changes requested):
Fix the comments. The GatewayRealm was probably a rebase-mistake. If everything passes, lets merge it.

As discussed, we might refactor it later with AgenticExposure and Subscription but for now, lets just merge it.

agentic/internal/handler/mcpserver/handler.go
```diff
@@ -0,0 +1,75 @@
...
+			candidates = append(candidates, serverList.Items[i])
+		}
+	}
+
+	// Determine active: oldest-wins semantics
+	sort.Slice(candidates, func(i, j int) bool {
+		return candidates[i].CreationTimestamp.Before(&candidates[j].CreationTimestamp)
+	})
+
+	if len(candidates) > 0 && candidates[0].UID != obj.UID {
```
ron96g: This could be improved. Check how API does it and try to use the same Conditions and messages. Its just nicer if this matches

agentic/internal/handler/mcpserver/handler.go
```diff
@@ -0,0 +1,75 @@
...
+	var candidates []agenticv1.McpServer
+	for i := range serverList.Items {
+		if serverList.Items[i].Spec.BasePath == obj.Spec.BasePath {
+			candidates = append(candidates, serverList.Items[i])
+		}
+	}
+
+	// Determine active: oldest-wins semantics
+	sort.Slice(candidates, func(i, j int) bool {
+		return candidates[i].CreationTimestamp.Before(&candidates[j].CreationTimestamp)
```
ron96g: Super edge case but still. Lets keep it similiar to API. Check the sortByCreationTime func there

agentic/internal/handler/mcpexposure/handler.go
```diff
@@ -0,0 +1,181 @@
...
+	// 5. Handle cross-zone proxy routes
+	obj.Status.Route = nil
+	obj.Status.ProxyRoutes = nil
+
+	crossZones, err := util.FindCrossZoneMcpSubscriptionZones(ctx, obj.Spec.BasePath, obj.Spec.Zone.Name)
+	if err != nil {
+		return errors.Wrap(err, "failed to find cross-zone MCP subscriptions")
+	}
+
+	for _, subscriberZoneRef := range crossZones {
```
ron96g, line 82: Question: What about DTC aka ConsumerFailover? Is this supported in the old CP?

agentic/internal/handler/mcpexposure/handler.go
```diff
@@ -0,0 +1,181 @@
...
+		proxyRoute, routeErr := util.CreateMcpProxyRoute(ctx, obj.Spec.BasePath, subscriberZone, zone)
+		if routeErr != nil {
+			return errors.Wrapf(routeErr, "failed to create MCP proxy Route for zone %q", subscriberZoneRef.Name)
+		}
+		obj.Status.ProxyRoutes = append(obj.Status.ProxyRoutes, *ctypes.ObjectRefFromObject(proxyRoute))
+		logger.V(1).Info("MCP proxy Route created/updated", "zone", subscriberZoneRef.Name, "route", proxyRoute.Name)
+	}
+
+	// 6. Resolve platform consumer for auto-access (TeleMCP variant)
+	telecontextConsumer := ""
```
ron96g: How exactly does this work? Maybe add docs to describe this in the Architecture section?

agentic/internal/handler/mcpexposure/handler.go
```diff
@@ -0,0 +1,181 @@
...
+	if obj.Spec.Variant.IsTelecontextVariant() {
+		if h.Config.TelecontextConsumerName == "" {
+			return errors.New("TELECONTEXTMCP variant requires telecontext consumer name to be configured")
+		}
+		telecontextConsumer = h.Config.TelecontextConsumerName
+	}
+
+	// 7. Create primary MCP route
+	isProxyTarget := len(obj.Status.ProxyRoutes) > 0
+	route, err := util.CreateMcpRoute(ctx, obj, zone, isProxyTarget, telecontextConsumer)
```
ron96g: What happens if the zone != telecontext-zone? Should this upport meshing?

agentic/internal/handler/mcpexposure/handler.go
```diff
@@ -0,0 +1,181 @@
...
+		return err
+	}
+
+	// 4. Check zone supports AI Gateway
+	if !zone.IsFeatureEnabled(adminv1.FeatureAiGateway) {
+		obj.SetCondition(condition.NewNotReadyCondition("AiGatewayNotSupported",
+			"Zone "+zone.Name+" does not support the AI Gateway feature"))
+		return ctrlerrors.BlockedErrorf("zone %q does not support the AI Gateway feature", zone.Name)
+	}
+
```
ron96g, line 71: In API there are more validation-checks. Is this not needed here?

agentic/internal/handler/mcpsubscription/handler.go
```diff
@@ -0,0 +1,310 @@
...
+	requester := &approvalapi.Requester{
+		TeamName:       requestorApp.Spec.Team,
+		TeamEmail:      requestorApp.Spec.TeamEmail,
+		ApplicationRef: types.TypedObjectRefFromObject(requestorApp, c.Scheme()),
+		Reason: fmt.Sprintf("Team %s requested MCP subscription to %s from zone %s",
+			requestorApp.Spec.Team, obj.Spec.BasePath, obj.Spec.Zone.Name),
+	}
+
+	properties := map[string]any{
+		"mcpBasePath": obj.Spec.BasePath,
```
ron96g, line 127: What about oauth2scopes? I have seen them in the CRs but its not something that we support in MCP?

agentic/internal/handler/mcpsubscription/handler.go
```diff
@@ -0,0 +1,310 @@
...
+	subscriberZone, err := util.GetZone(ctx, obj.Spec.Zone.K8s())
+	if err != nil {
+		return err
+	}
+	if !subscriberZone.IsFeatureEnabled(adminv1.FeatureAiGateway) {
+		obj.SetCondition(condition.NewNotReadyCondition("AiGatewayNotSupported",
+			"Subscriber zone "+subscriberZone.Name+" does not support the AI Gateway feature"))
+		return ctrlerrors.BlockedErrorf("subscriber zone %q does not support the AI Gateway feature", subscriberZone.Name)
+	}
+
```
ron96g, line 96: scopes validation and potentially other validation-checks that are in API might be missing

agentic/internal/handler/mcpsubscription/handler.go
```diff
@@ -0,0 +1,310 @@
...
+			Security:     util.MapSubscriberSecurityToGateway(obj.Spec.Security),
+			Traffic:      util.MapSubscriberTrafficToGateway(&obj.Spec.Traffic),
+		}
+		return nil
+	}
+
+	_, err := c.CreateOrUpdate(ctx, consumeRoute, mutator)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil, ctrlerrors.BlockedErrorf("Route %q not found — McpExposure may not have created the proxy route yet", routeRef.String())
```
ron96g: It does not work like this

ron96g (changes requested):
- how exactly does telecontext work? Where is telecontext running? Do we need to mesh it?
- we copied all fields from API but I dont see the validation for it. Also Approval must respect scopes if we support scopes --> What features are actually supported?
- There is a big difference between old and new CP. In the old the uniqueness of the basePath was ensured via a shared API CR. Now we have API and McpServer. How do we ensure that they isnt a conflict for the same basePath?
-

agentic/internal/handler/mcpexposure/handler.go
julius-malcovsky: improved - checking for basepath case sensitivity. also added proxy route cleanup.

scopes are a bit of a problem - lets discuss them separately

api categories could also be added, but there are related issues - lets talk

agentic/internal/handler/mcpexposure/handler.go
julius-malcovsky: in the old CP yes, because it was a part of the original stargate domain so unless it was explicitly excluded it supported failover "naturally" ... in this first version of agentic its not yet supported - i would propose of doing it in the next iteration, this PR is already big enough - what do you think ?

agentic/internal/handler/mcpexposure/handler.go
julius-malcovsky: added a file in docs

agentic/internal/handler/mcpsubscription/handler.go
julius-malcovsky: resolveRouteRef added as a simplified version of the api domains func (doesnt include the failover logic yet)

rover/api/v1/rover_types.go
julius-malcovsky: analyzed it and it should - its technically a union style struct and exactly 1 is accepted

julius-malcovsky: It works because the type fields are  omitempty  pointers — only the one you set appears in the serialized JSON, so min/max properties = 1 enforces exactly one type is present.

admin/internal/handler/zone/handler.go
```diff
@@ -66,6 +74,16 @@ func (h *ZoneHandler) CreateOrUpdate(ctx context.Context, obj *adminv1.Zone) err
 		return nil
 	}
 
+	// AI Gateway configuration
+	if cconfig.FeatureAiGateway.IsEnabled() && obj.Spec.AiGateway != nil {
```
ron96g: this breaks the step-function pattern. Needs to be refactored to match it

agentic/internal/controller/mcpexposure_controller.go
```diff
@@ -0,0 +1,232 @@
...
+// +kubebuilder:rbac:groups=agentic.cp.ei.telekom.de,resources=mcpexposures/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=agentic.cp.ei.telekom.de,resources=mcpexposures/finalizers,verbs=update
+// +kubebuilder:rbac:groups=admin.cp.ei.telekom.de,resources=zones,verbs=get;list;watch
+// +kubebuilder:rbac:groups=admin.cp.ei.telekom.de,resources=zones/status,verbs=get
+// +kubebuilder:rbac:groups=gateway.cp.ei.telekom.de,resources=routes,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=gateway.cp.ei.telekom.de,resources=realms,verbs=get;list;watch
+// +kubebuilder:rbac:groups=gateway.cp.ei.telekom.de,resources=consumeroutes,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=agentic.cp.ei.telekom.de,resources=mcpservers,verbs=get;list;watch
+// +kubebuilder:rbac:groups=agentic.cp.ei.telekom.de,resources=mcpsubscriptions,verbs=get;list;watch
+// +kubebuilder:rbac:groups=identity.cp.ei.telekom.de,resources=clients,verbs=get;list;watch
```
ron96g: are all these needed? For what is Client used here?

agentic/internal/controller/predicates.go
```diff
@@ -0,0 +1,30 @@
...
+
+	"github.com/telekom/controlplane/common/pkg/config"
+)
+
+const domainName = "agentic"
+
+// LabelPredicate filters events for resources with domain=agentic label.
+var LabelPredicate = predicate.Funcs{
+	CreateFunc: func(e event.CreateEvent) bool {
+		return e.Object.GetLabels()[config.DomainLabelKey] == domainName
```
ron96g, line 19: can GetLabels return nil?

agentic/internal/handler/util/route.go
```diff
@@ -0,0 +1,416 @@
...
+				DisableResponseBuffering: true,
+			},
+		}
+
+		// Apply security settings
+		if exposure.Spec.Security != nil {
+			route.Spec.Security = MapSecurityToGateway(exposure.Spec.Security)
+		} else {
+			route.Spec.Security = gatewayapi.Security{
+				DisableAccessControl: true,
```
ron96g: is this correct?

ron96g (approved):

PR #434, opened 2026-06-19
feat(projector-expired-filter): internal state Expired is filtered by

julius-malcovsky:
The 'Expired' state of an approval is internal and has special logic. Currently we do not want to make the transition into Expired visible to the CP API consumers and thus projector filters it out


BjoernKarma (approved):
LGTM

PR #435, opened 2026-06-22
feat: initial commit for restructuring of admin and gateway

ron96g:
This PR completely restructured the Gateway and Admin-Domain:
- It removes the logical Gateway-Realm CR to align more with the Kubernetes-Gateway-API
- It therefore introduces Gateway-Presets that are configured per zone and enable similar functionality.
- It restructures the Gateway-Route CR to now accept all configuration explicitly and no longer implicitly merged anything in the gateway-operator
- Due to the central nature of the Admin-Zone and Gateway-{Route, Realm} CRs all other domains which consumed them had to be updated accordingly.
- ConsumerFailover is now wired into all relevant domains (rover, application, api and gateway)
- ProviderFailver has been improved and now allow to configure multiple zones (instead of just array of size 1)

Additional Features:
- Improves the bootstrapping of global resources in the Zone-Handlers by using (A) step-functions and (B) secret management
  - (A) step-functions allow us to extend this naturally and also keeps the handler maintainable
  - (B) secret-management now automatically replaces cleartext-secrets with their secret-refs in the zone-webhook
- Remove the Gateway-Client used to meshing in favour of new LMS-Meshing offered in the newest Jumper version
- Bump the envtest-version to 1.32.0 due to `dns1123Subdomain` format validation


BjoernKarma (approved):
LGTM as discussed in the team review

PR #438, opened 2026-06-25
fix(secret-manager): harden k8s onboarder

ron96g:


BjoernKarma (dismissed):
LGTM - The copilot review comments are for the test. We could fix that, but also merging is fine since the fix has proven to do what it is supposed to.

BjoernKarma (approved):
LGTM - I really like the contract-based testing for the secret-manager backends

PR #446, opened 2026-06-30
ci: splitting the single gomod Dependabot update definition

BjoernKarma:
splitting the single gomod Dependabot update definition into 4 smaller gomod update blocks 
(all still monthly, all still using the same minor-patch grouping)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -21,34 +21,61 @@ updates:
       - application/api
       - approval
       - approval/api
+      - cpapi/api
+    groups:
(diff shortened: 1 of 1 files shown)


stefan-ctrl (commented):
Was this splitting done based on any specific grouping?

I kinda fear, since we use `replace`-statements, that having different jobs would diverge the `go.sum` version entries, resulting in broken builds.

stefan-ctrl (approved):
Discussed off-github. Sounds good and will evaluate this approach

PR #447, opened 2026-06-30
ci: split ci into a matrix job 

BjoernKarma:
### What changed

The `ci.yaml` workflow previously defined 22 separate top-level jobs — one per module — each calling `reusable-go-ci.yaml` individually. This PR replaces that repetitive structure with:

- A single `go_ci` **matrix job** covering 19 modules (up to 6 running in parallel), with all per-module configuration encoded as matrix entries.
- Two dedicated direct jobs (`common`, `common_server`) for modules with non-standard flags (e.g. `run_build_image: false` for the library-only `common` module).
- One dedicated `snapshotter` job because it requires `run_lint: false`, which cannot be expressed in the shared matrix.
- A new `ci_summary` job that aggregates all results and acts as a single aggregation point for branch protection rules.

### Why this makes sense

| Before | After |
|---|---|
| 22 individually written jobs | 1 matrix + 3 special-case jobs |
| Adding a module = copy-pasting ~10 lines | Adding a module = 4–5 line matrix entry |
| All jobs visible as separate top-level status checks | Single `CI Summary` check aggregates everything |
| One failure could obscure others | `fail-fast: false` — all modules run, all failures visible |

### What is preserved

- **All per-module flags** (`coverage_threshold`, `run_check_generated_files`, `lint_fail_on_issues`, `ko_build_path`) are encoded in the matrix entries with the same values as before.
- **Internal test-before-build ordering** inside `reusable-go-ci.yaml` is unchanged — each call still runs `tests → build image`.
- **Helm release path** (`common_server_helm`) still depends on `common_server` and only runs on tag pushes.
- **No behavior change** on PR, push to `main`, tag, schedule, or `workflow_dispatch` triggers.

### Action required after merge

If branch protection rules reference specific job names from the old workflow (e.g. `"Approval / Tests & Coverage for approval"`), update them to reference **`CI Summary`** as the single required status check.


BjoernKarma:
Superseded by https://github.com/telekom/controlplane/pull/626

PR #454, opened 2026-07-01
feat(cpapi): security hardening, disable mock auth, add JWT

iagarba:


ron96g (commented):
What is the benefit of this structure instead of just using the trustedIssuers as the determining factor like we do in common-server?

Example:
If len(trustedIssuers) == 0 --> Mock-Mode
if len(trustedIssuers) > 0 --> Prod-Mode

I am not sure if we would ever disable it

controlplane-api/cmd/config/config.go
```diff
@@ -43,6 +44,9 @@ type TLSConfig struct {
 }
 
 type SecurityConfig struct {
+	// Mode controls authentication behaviour: disabled, mock, or jwt.
+	// Deprecated: Enabled is derived from Mode and retained for backward compatibility.
+	Mode           string   `yaml:"mode"`
 	Enabled        bool     `yaml:"enabled"`
 	TrustedIssuers []string `yaml:"trustedIssuers"`
 }
```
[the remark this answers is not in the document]
iagarba: check it again

controlplane-api/cmd/main.go
```diff
@@ -147,6 +151,19 @@ func main() {
 	}
 }
 
+// validateSecurityConfig logs the active security mode at startup.
+// Structural validation (jwt requires trusted issuers) is enforced by ConfigureSecurity.
+func validateSecurityConfig(log logr.Logger, sec config.SecurityConfig) {
```
ron96g: Why put it here and not in common-server?

ron96g: Also does not really validate but logs what mode is selected

controlplane-api/cmd/config/config.go
```diff
@@ -61,6 +65,23 @@ type FileManagerConfig struct {
 	BaseURL string `yaml:"baseUrl"`
 }
 
+// Validate checks the security configuration for invalid or unsafe combinations.
+// It returns an error for unknown modes or for jwt mode without trusted issuers.
+// Call this at startup and panic on error to implement fail-closed behaviour.
+func (sec SecurityConfig) Validate() error {
```
ron96g: Is this used anywhere?

common-server/pkg/server/middleware/security/security.go
```diff
@@ -19,51 +21,61 @@ const (
...
-		opts.Log.Info("⚠️ Security middleware mocked")
-		router.Use(mock.NewJWTMock())
-	} else {
+	switch opts.Mode {
+	case ModeJWT:
+		jwtOpts := &JWTOpts{}
+		for _, f := range opts.JWTOpts {
+			f(jwtOpts)
+		}
+		if len(jwtOpts.TrustedIssuers) == 0 {
```
ron96g, line 64: Why not used the Validate function here? Or delete the Validate function

ron96g (changes requested):

controlplane-api/cmd/main.go
iagarba: leftover from first implementation I guess

controlplane-api/cmd/config/config.go
iagarba: dead code, will be rmoved

common-server/pkg/server/middleware/security/security.go
iagarba: removed Validate

ron96g (dismissed):

ron96g (approved):

PR #466, opened 2026-07-06
fix: improve failover handling; add mesh-flag to routing-config

ron96g:
This PR improve the handling of ConsumerFailover (CF) in the api module. It now checks for each Route if the zone supports the feature, e. g. it could be that the API is exposed on a Zone that does not support CF, however, the subscriberZone and others do support it. Then CF must be configured on those routes but cannot be configured on the primary-route.

### DIFF
--- a/api/internal/handler/apiexposure/handler_test.go
+++ b/api/internal/handler/apiexposure/handler_test.go
@@ -168,10 +168,6 @@ var _ = Describe("ApiExposureHandler", func() {
 			return makeReadyZone("zone-b", "ns-b",
 				"https://idp.zone-b.example.com", "https://lms.zone-b.example.com")
 		}
-		zoneC = func() *adminv1.Zone {
-			return makeReadyZone("zone-c", "ns-c",
-				"https://idp.zone-c.example.com", "https://lms.zone-c.example.com")
-		}
 		zoneF = func() *adminv1.Zone {
 			return makeReadyZone("zone-f", "ns-f",
 				"https://idp.zone-f.example.com", "https://lms.zone-f.example.com")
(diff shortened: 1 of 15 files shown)


rover/internal/webhook/v1/rover_webhook.go
```diff
@@ -121,6 +121,15 @@ func (r *RoverValidator) ValidateCreateOrUpdate(ctx context.Context, rover *rove
...
 	}
 
+	// Validate ConsumerFailover: feature must be enabled on this zone if ConsumerFailover is configured
+	if !zone.IsFeatureEnabled(adminv1.FeatureConsumerFailover) && rover.HasFailoverEnabledOnAnySubscription() {
+		valErr.AddInvalidError(
+			field.NewPath("spec").Child("zone"),
+			rover.Spec.Zone,
+			fmt.Sprintf("zone %q does not support consumer failover. Either disable it or select a different zone", rover.Spec.Zone))
+		return nil, valErr.BuildError()
+	}
```
stefan-ctrl, line 131: Minor: Would it be possible to add a test for that in `rover_webhook_test.go`?

api/internal/handler/apiexposure/handler_test.go
stefan-ctrl, line 1: Minor:

It would be nice to have tests that check for the blocked state of resources if not allowed states are within the cluster (will unlikely to happen due to checks at the rover level, add tests for being sure that it works as intended)

https://github.com/telekom/controlplane/blob/7e203374925705da1364c8f163dc8f8dc628a776/api/internal/handler/apiexposure/handler.go#L564-L578

stefan-ctrl (approved):
LGTM.

Have two minor comments about test coverage, but should not block this PR from being merged in my opinion

PR #468, opened 2026-07-07
feat(event): add category to all event resources

ron96g:
This is currently just a string and could be later extended to a similiar pattern as in api

### DIFF
--- a/event/api/v1/eventtype_types.go
+++ b/event/api/v1/eventtype_types.go
@@ -43,13 +43,11 @@ type EventTypeSpec struct {
 	// +optional
 	Description string `json:"description,omitempty"`
 
-	// Category of the event, defaults to "other" if not specified.
-	// Restricted to lowercase letters to keep a future migration to a
-	// dedicated EventCategory CR straightforward.
-	// +kubebuilder:validation:Required
+	// Category of the event.
+	// +optional
 	// +kubebuilder:validation:Pattern=`^[a-z]+$`
 	// +kubebuilder:default:=other
-	Category string `json:"category"`
+	Category string `json:"category,omitempty"`
 
 	// Specification contains the file ID reference from the file manager for
 	// the optional JSON schema that describes the event payload.
--- a/event/api/v1/eventtype_types.go
+++ b/event/api/v1/eventtype_types.go
@@ -44,7 +44,10 @@ type EventTypeSpec struct {
 	Description string `json:"description,omitempty"`
 
 	// Category of the event, defaults to "other" if not specified.
+	// Restricted to lowercase letters to keep a future migration to a
+	// dedicated EventCategory CR straightforward.
 	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Pattern=`^[a-z]+$`
 	// +kubebuilder:default:=other
 	Category string `json:"category"`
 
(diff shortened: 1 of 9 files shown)


event/api/v1/eventtype_types.go
```diff
@@ -43,12 +43,15 @@ type EventTypeSpec struct {
 	// +optional
 	Description string `json:"description,omitempty"`
 
+	// Category of the event, defaults to "other" if not specified.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:default:=other
+	Category string `json:"category"`
```
stefan-ctrl: Do we want to add a couple more limitations until a EventCategory CR is implemented?

I'm thinking about a regex that restricts the `Category string` to valid k8s name chars and enforces lowercase to make later migration easier.

stefan-ctrl (changes requested):

event/api/v1/eventtype_types.go
ron96g: Good idea. We could check for `[a-z]+` and then in rover-server cast it to lower-case. I dont think there are any EventSpecs in the old CP where we have numbers - might need to check

rover/internal/handler/eventspecification/handler.go
```diff
@@ -47,10 +47,18 @@ func (h *EventSpecificationHandler) CreateOrUpdate(ctx context.Context, eventSpe
 			eventv1.EventTypeLabelKey: labelutil.NormalizeLabelValue(eventSpec.Spec.Type),
 		}
 
+		category := eventSpec.Spec.Category
```
BjoernKarma: Should we add a Gingko/Gomega test for the fallback?

event/api/v1/eventtype_types.go
```diff
@@ -43,12 +43,18 @@ type EventTypeSpec struct {
 	// +optional
 	Description string `json:"description,omitempty"`
 
+	// Category of the event, defaults to "other" if not specified.
+	// Restricted to lowercase letters to keep a future migration to a
+	// dedicated EventCategory CR straightforward.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Pattern=`^[a-z]+$`
+	// +kubebuilder:default:=other
+	Category string `json:"category"`
```
BjoernKarma: Do we need the "omitempty" here as well? To handle existing eventtypes that are then updated with the CRD fallback? Or do we want to update all existing eventtypes? 
I think this might be related to the test error: https://github.com/telekom/controlplane/actions/runs/29008022924/job/86084335567?pr=468

rover/api/v1/eventspecification_types.go
```diff
@@ -37,6 +37,14 @@ type EventSpecificationSpec struct {
 	// +optional
 	Description string `json:"description,omitempty"`
 
+	// Category of the event, defaults to "other" if not specified.
+	// Restricted to lowercase letters to keep a future migration to a
+	// dedicated EventCategory CR straightforward.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Pattern=`^[a-z]+$`
+	// +kubebuilder:default:=other
+	Category string `json:"category"`
```
BjoernKarma: See comment above for eventtypes

BjoernKarma (commented):
I have added two questions:
- Using `omitempty` for `Category`?
- Adding a unit test for the fallback with `other`?

stefan-ctrl (approved):
LGTM

PR #469, opened 2026-07-08
feat(event): support proxy-zones

ron96g:
- **fix(event): correct publish paths**
- **feat(event): add support for event proxy-zones; refactored route-creation; added detailed docs for event-routes**


docs/docs/admin-journey/environments-and-zones.md
```diff
@@ -266,25 +283,55 @@ mesh:
...
+    namespace: dev
+  proxy:
+    targetZone:
+      name: dataplane1
       namespace: dev
 ```
 
+The target zone must be a **local** (non-proxy) zone running Horizon. Because a proxy zone has no backend of its own, it needs no `admin` or backend URLs — those all belong to the target zone.
+
+:::warning
```
julius-malcovsky: this does not seem to render properly in the docs - view the docs file 

Image

event/internal/handler/eventconfig/handler.go
```diff
@@ -394,6 +481,82 @@ func (h *EventConfigHandler) createVoyagerRoutes(ctx context.Context, obj *event
...
+	logger := log.FromContext(ctx)
+
+	realmName := myZone.Status.RealmName
+	targetZoneName := obj.Spec.Proxy.TargetZone.Name
+
+	targetCfg, err := util.GetEventConfigForZone(ctx, targetZoneName)
+	if err != nil {
+		return err // BlockedError propagates so the proxy requeues until the target is ready
+	}
+	if targetCfg.IsProxy() || targetCfg.Spec.Local == nil {
```
julius-malcovsky: (possible nitpick) does an event config has to have at least a local or a proxy config ? if yes then maybe this could be simplified - it feels like were asking "if nothing is configured" ... if we checked for that elsewhere we wouldnt have to re-validate it here ... what do you think?

event/internal/handler/eventconfig/handler.go
```diff
@@ -86,17 +99,31 @@ func (h *EventConfigHandler) CreateOrUpdate(ctx context.Context, obj *eventv1.Ev
 	}
 	logger.V(1).Info("Callback Routes created/updated", "count", len(obj.Status.ProxyCallbackRoutes))
 
-	if obj.Spec.VoyagerApiUrl != "" {
+	// Voyager Routes: local zones expose their own backend; proxy zones forward their
+	// own-zone Route to the target and still participate in the full mesh.
+	if obj.IsProxy() {
+		if err = h.createProxyVoyagerRoutes(ctx, obj, myZone, meshCfg); err != nil {
```
julius-malcovsky: cosmetic - felt like 2 different conditions are checked here .. either is proxy .. then we dont check for voyagerApiUrl ELSE we know its local but this time we check if apiUrl != ""  ..

event/internal/handler/eventconfig/handler.go
```diff
@@ -428,6 +591,120 @@ func (h *EventConfigHandler) createEventStore(ctx context.Context, obj *eventv1.
...
+func (h *EventConfigHandler) createProxyEventStore(ctx context.Context, obj *eventv1.EventConfig) (*pubsubv1.EventStore, error) {
+	c := cclient.ClientFromContextOrDie(ctx)
+
+	targetZoneName := obj.Spec.Proxy.TargetZone.Name
+
+	targetCfg, err := util.GetEventConfigForZone(ctx, targetZoneName)
+	if err != nil {
+		return nil, err // BlockedError propagates so the proxy requeues until the target is ready
+	}
+	if targetCfg.IsProxy() || targetCfg.Spec.Local == nil {
```
julius-malcovsky: maybe could be simplified? we have this helper func IsProxy(), we could have IsLocal() too and I think it would be enough to ask if "!IsLocal()" .. because the CEL validation should reject CRs that would have invalid config anyways .. got this feeling were doublechecking a lot + the readability suffers a little - helper func vs "manual check"

event/internal/handler/eventconfig/handler.go
```diff
@@ -428,6 +591,120 @@ func (h *EventConfigHandler) createEventStore(ctx context.Context, obj *eventv1.
 	return eventStore, nil
 }
 
+// createProxyEventStore creates a pubsub.EventStore for a proxy zone. Instead of a local
+// admin client, it authenticates to the target zone's configuration backend using the
+// target zone's admin client credentials and admin realm token endpoint.
+func (h *EventConfigHandler) createProxyEventStore(ctx context.Context, obj *eventv1.EventConfig) (*pubsubv1.EventStore, error) {
```
julius-malcovsky, line 612: feel free to resolve this, its just for my understanding - why is it necessary to create an event store for a different zone? I must be missing something, but I thought that the eventConfig either is present for that zone and then the eventStore is also created or it doesnt make sense to have an eventStore there anyways ... ill keep reading, maybe ill find the answer :) maybe a short line in the comments would be nice - we do this because (ill make this up) its needed for the proxy callback routes ..

event/internal/handler/eventconfig/handler.go
```diff
@@ -428,6 +591,120 @@ func (h *EventConfigHandler) createEventStore(ctx context.Context, obj *eventv1.
...
+
+	realmName := myZone.Status.RealmName
+
+	// Publishers access the proxy publish route with IDP tokens, same as a primary route.
+	var trustedIssuers []string
+	if myZone.Status.Links.Issuer != "" {
+		trustedIssuers = []string{myZone.Status.Links.Issuer}
+	}
+
+	route, err := util.CreatePublishProxyRoute(ctx, myZone, targetZone,
```
julius-malcovsky, line 709: (ubernitpick) CreateProxyPublisRoute vs CreatePublishProxyRoute - maybe pick one? different packages so you should be able to use the same name :)

event/internal/handler/eventexposure/handler.go
```diff
@@ -133,29 +134,67 @@ func (h *EventExposureHandler) reconcileSSERoutes(ctx context.Context, obj *even
...
 	crossZones, err := util.FindCrossZoneSSESubscriptionZones(ctx, obj.Spec.EventType, obj.Spec.Zone.Name)
 	if err != nil {
 		return errors.Wrap(err, "failed to find cross-zone SSE subscriptions")
 	}
 
-	subscriberZones, err := h.createProxySSERoutes(ctx, obj, zone, crossZones, realmName)
+	// Subscriber proxy Routes forward to the backend zone (not the exposure zone,
+	// which may itself be a proxy). The backend zone is served directly by the
+	// primary Route below, so it is skipped inside createProxySSERoutes.
+	subscriberZones, err := h.createProxySSERoutes(ctx, obj, backendZone, crossZones, realmName)
```
julius-malcovsky, line 154: (nitpick ... again) ... might be a bit confusing, the func is called create .. Routes and it returns zones?

event/internal/handler/util/callback_route.go
```diff
@@ -6,101 +6,36 @@ package util
...
 	gatewayapi "github.com/telekom/controlplane/gateway/api/v1"
 )
 
-//nolint:dupl // parallel structure with CreateProxyVoyagerRoute; differs in naming, labels, and security
+// How this works:
+//
+// Subscriber and Provider share the same Zone ("provider" --> "foo" --> "consumer"):
+// 1. Subscriber has deliveryType callback with the callbackUrl "https://my-callback/v1/post"
+// 2. Approval is done and pubsub-Subscription is provisioned with the callbackUrl "https://foo-gateway/horizon-foo/callback/v1?callback=https://my-callback/v1/post"
+// 3. Horizon publishes events (auth using mesh-client) which will be proxies to the foo-gateway and then forwarded the the callbackUrl with the correct LMS-token (clientId=mesh-client)
```
julius-malcovsky: "publishes events which will be proxies to the foo-gateway" - can you explain this using other words perhaps? It sounds like events are proxies to the gateway ... not sure how to understand this

event/internal/handler/util/callback_route.go
```diff
@@ -171,80 +98,20 @@ func CreateCallbackRoute(
...
-
-	var collected []*adminv1.Zone
-	for _, zone := range candidates {
-		if _, ok := wantedSet[zone.Name]; ok {
-			collected = append(collected, zone)
-		}
-	}
-
-	return collected
+	return createProxyRoutes(ctx, meshConfig, sourceZone, targetZones, CreateProxyCallbackRoute, opts...)
```
julius-malcovsky, line 116: Perhaps we could think of using a Parameter object? the list of params is starting to resemble the old gateway ...

event/internal/handler/util/shared_route.go
```diff
@@ -129,3 +148,189 @@ func RouteDownstreamURL(route *gatewayapi.Route) string {
...
+	for _, opt := range opts {
+		opt(options)
+	}
+
+	sourcePreset, err := resolvePreset(sourceZone)
+	if err != nil {
+		return nil, err
+	}
+
+	targetPreset, err := targetZone.Spec.Gateway.GetDefaultPreset()
```
julius-malcovsky: would it make sense to call the resolvePreset func here ?

julius-malcovsky: out of scope now but i think we could create a preset resolver, think we already discussed it partially - request a set of features and get a preset that can handle it / error .. i guess the default preset could be an exception, but otherwise its a nice feature, would be nice to have some convenience funcs so you dont manually search for

event/internal/handler/util/sse_route.go
```diff
@@ -21,6 +20,18 @@ import (
...
+// How this works:
+//
+// Subscriber and Provider share the same Zone ("consumer" --> "foo" <-- "provider"):
+// 1. Subscriber has deliveryType sse
+// 2. Approval is done and Subscriber gets assigned the sseUrl "https://foo-gateway/horizon/sse/v1/<eventType>/<subscriptionId>"
+// 3. Subscriber connect via the sseUrl to the foo-gateway which proxies the connection to Horizon and then forwards the events to the subscriber
+//
+// Subscriber is on a different zone ("consumer" --> "bar" --> "foo" <-- "provider"):
+// 1. Subscriber has deliveryType sse
+// 2. Approval is done and Subscriber gets assigned the sseUrl "https://bar-gateway/horizon/sse/v1/<eventType>/<subscriptionId>" (same path as the eventType is a singleton in the system)
```
julius-malcovsky, line 32: wonder if it would make sense to be consistent with callback style https://bar-gateway/horizon-foo/callback/v1?callback=https://my-callback/v1/post" --- i mean the /horizon-foo segment ... i guess we are also limited by horizon itself and this might be just an example you wrote so i dont take it that seriously .. but i liked the info value of the url itself .. im a gateway on bar and i route traffic from foo ... pretty sweet :)

julius-malcovsky (dismissed):
nicely done, i found some small things that might we worth taking a look at but nothing major as far as I can tell - if you find the time, please answer the questions I wrote or contact me directly, would be happy to have a call about those points. 
Thanks for your contribution, approved :)

docs/docs/admin-journey/environments-and-zones.md
ron96g: fixed

event/internal/handler/eventconfig/handler.go
ron96g: improved

event/internal/handler/eventconfig/handler.go
ron96g: fixed

event/internal/handler/eventconfig/handler.go
ron96g: fixed

event/internal/handler/eventconfig/handler.go
ron96g: Its basically just there to satisfy the existing logic. It points to the target-zone but if inside of the CP its still its own instance

event/internal/handler/util/callback_route.go
ron96g: improved readability

julius-malcovsky (approved):
all good now

PR #471, opened 2026-07-09
fix: we must not rely on readiness for uniqueness check in the singleton logic

ron96g:
We cannot use the readiness-check for singleton selection. If - for whatever reason - an Api, ApiExposure, EventType or EventExposure becomes NOT READY, it would result in a singleton switch


stefan-ctrl (dismissed):

BjoernKarma (approved):
LGTM - as discussed in team review

PR #472, opened 2026-07-09
fix(projector): back-link subscriptions to fix race-condition exposure vs subscription

ron96g:
Subscriptions that were projected before the targeted Exposure are allowed (Blocked until exposure is ready).
However, the back-link was never set afterwards. Therefore the Subscriptions remains without a target.


stefan-ctrl (approved):
LGTM.

`xxx.Not(apisubscription.HasTarget()),`  should prevent some kind of cyclic triggering.
If so, feel free to merge.

PR #485, opened 2026-07-12
feat: add organization-server

julius-malcovsky:
## Summary

This PR introduces the **Organization Server**, a compatibility facade that exposes the existing Hub and Team REST API while using the new Control Plane architecture underneath.

The service translates legacy organization-management requests into Control Plane GraphQL operations and retrieves team-owned resources from Rover Server. This allows existing clients to migrate without changing their current API contract.

## What Changed

- Added the new `organization-server` component, including:
  - OpenAPI specification and generated server code
  - Hub and Team management endpoints
  - Team token rotation
  - Team resource inventory
  - Status endpoints
  - Kubernetes deployment, RBAC, ingress, TLS, and monitoring configuration
- Added Hub and Team mutations to Control Plane API:
  - Create, update, and delete groups
  - Create, update, and delete teams
  - Add and remove team members
  - Rotate team tokens
- Added deletion safeguards:
  - Teams cannot be deleted while they still own resources
  - Hubs cannot be deleted while they still contain teams
- Added a paginated Rover Server `/resources` endpoint that:
  - Aggregates all resources owned by a team
  - Supports team, group, and admin access scopes
  - Covers Rovers, API specifications, event specifications, roadmaps, API changelogs, and MCP specifications
- Added JWT authentication for external requests and projected Kubernetes service-account authentication for internal communication.
- Added TLS-secured communication between Organization Server, Control Plane API, and Rover Server.
- Preserved sensitive-data obfuscation according to the caller's access scope.
- Added build, release, CI, and local-development integration for the new component.
- Fixed Projector startup processing so existing resources are handled correctly after restart.

## Architecture

```text
Legacy client
    |
    v
Organization Server
    |
    +--> Control Plane API (Hub and Team operations)
    |
    +--> Rover Server (team resource inventory)
```

The Organization Server keeps the legacy REST representation while delegating persistence and business operations to the existing Control Plane services.

## Testing

Added coverage for:

- Hub and Team operations
- Request and response mapping
- Authentication and authorization
- Sensitive-data obfuscation
- Upstream client failures and error responses
- Team and Hub deletion safeguards
- Resource filtering and pagination
- Pagination across more than 500 resources
- Invalid and non-advancing cursors
- TLS and configuration validation

CI checks are currently running; completed static checks, Organization Server tests, vulnerability scans, and documentation builds are passing.


controlplane-api/internal/service/resource_checker.go
```diff
@@ -0,0 +1,89 @@
...
+	req.Header.Set("Authorization", "Bearer "+token)
+
+	resp, err := r.httpClient.Do(req)
+	if err != nil {
+		return false, fmt.Errorf("calling rover-server: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return false, fmt.Errorf("rover-server returned status %d", resp.StatusCode)
```
ron96g: improve error handling

controlplane-api/internal/service/services_k8s.go
```diff
@@ -234,6 +235,60 @@ func (s *teamK8sService) RotateTeamToken(ctx context.Context, ref ResourceRef) (
...
+				Message: "unable to verify team resources, please try again",
+			}},
+		}, nil
+	}
+	if hasResources {
+		log.V(1).Info("Team still has resources, refusing deletion")
+		return &model.DeleteTeamPayload{
+			Errors: []model.MutationError{{
+				Code:    model.ErrorCodePreconditionFailed,
+				Message: "cannot delete team: team still has resources — delete all resources first",
```
ron96g, line 264: This feelsl ike a downgrade to the current solution where a user gets a list of resources that are blocking the deletion

controlplane-api/internal/service/services_k8s.go
```diff
@@ -234,6 +235,60 @@ func (s *teamK8sService) RotateTeamToken(ctx context.Context, ref ResourceRef) (
...
+			}},
+		}, nil
+	}
+
+	team := &organizationv1.Team{}
+	if err := s.client.Get(ctx, k8stypes.NamespacedName{Name: ref.Name, Namespace: ref.Namespace}, team); err != nil {
```
ron96g, line 270: Should ignore NotFound

rover-server/internal/controller/resources.go
```diff
@@ -0,0 +1,135 @@
...
+		return fmt.Errorf("listing %s: %w", rk.kind, err)
+	}
+
+	for _, item := range objList.Items {
+		ref := api.ResourceRef{
+			ApiVersion: rk.apiVersion,
+			Kind:       rk.kind,
+			Name:       item.GetName(),
+			Namespace:  item.GetNamespace(),
+			Path:       fmt.Sprintf("%s/%s", rk.pathPrefix, item.GetName()),
```
ron96g: This will not work as expected due to the proxying of Stargate, e.g. User submited `/eni/rover/v3` and we listen on `rover/api` so we either need to get the downstream value or this will be invalid

rover-server/internal/controller/resources.go
```diff
@@ -0,0 +1,135 @@
...
+	{apiVersion: roverv1.GroupVersion.String(), kind: "Roadmap", pathPrefix: "/apiroadmaps"},
+	{apiVersion: roverv1.GroupVersion.String(), kind: "ApiChangelog", pathPrefix: "/apichangelogs"},
+}
+
+func (r *ResourcesControllerImpl) GetAll(ctx context.Context, params api.GetAllResourcesParams) (*api.ResourceListResponse, error) {
+	tokenPrefix := security.PrefixFromContext(ctx)
+
+	// Determine effective prefix: use the explicit query param if provided,
+	// otherwise fall back to the token-derived prefix.
+	effectivePrefix, err := resolvePrefix(tokenPrefix, params.Prefix)
```
ron96g: This functions should only work on team-level. So we expect that there is always a `Team-Ref` provided by the user - even if its an admin or group scope.  Therefore I would propse to rename the `prefix` param to `group` and `team` and enforce them.

organization-server/internal/client/token.go
```diff
@@ -0,0 +1,87 @@
...
+	"net/http"
+	"net/url"
+	"strings"
+	"sync"
+	"time"
+)
+
+// TokenSource provides an OAuth access token, refreshing it automatically
+// when expired via the client_credentials grant.
+type TokenSource struct {
```
ron96g: Prefer `golang.org/x/oauth2/clientcredentials`

organization-server/internal/client/rover.go
```diff
@@ -0,0 +1,87 @@
...
+func (r *RoverClient) GetResources(ctx context.Context, environment, group, team string) (*ResourceListResponse, error) {
+	prefix := fmt.Sprintf("%s--%s--%s/", environment, group, team)
+	url := fmt.Sprintf("%s/resources?prefix=%s", r.baseURL, prefix)
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, http.NoBody)
+	if err != nil {
+		return nil, fmt.Errorf("building request: %w", err)
+	}
+
+	token := mock.NewMockAccessToken(environment, "org-server", "service", []string{r.scopePrefix + ":admin:all"})
```
ron96g: Hardcoded mock token?

organization-server/internal/handler/hubs.go
```diff
@@ -0,0 +1,241 @@
...
+	ctx := h.contextWithIdentity(c)
+	id := mw.ConsumerIdentityFromContext(c)
+	desc := req.Description
+	resp, err := gql.CreateGroup(ctx, h.cpapi, gql.CreateGroupInput{
+		Environment: id.Environment,
+		Name:        req.Name,
+		DisplayName: req.DisplayName,
+		Description: &desc,
+	})
+	if err != nil {
```
ron96g: better error handling

ron96g (changes requested):
first review. Looks mostly good

controlplane-api/internal/service/services_k8s.go
julius-malcovsky: are you sure about this? based on the oas and the code ive checked we did not return the list, we just told the customer to go check that endpoint (/resources)

organization-server/internal/client/rover.go
julius-malcovsky: yes I am aware this is far from ideal - the idea was that the rover-server currently runs with mock and we wanted to refactor it to a proper solution for all internal comms so i made this shortcut, expecting it to be short lived anyway. would it help to make it configurable, or did you ahve something else in mind

rover-server/internal/controller/resources.go
julius-malcovsky: we could get that from the managed routes of the zone right? but rover-server doesnt have RBAC to read zones .. as an alternative we could put that value info a config file, but then its not coupled if it changes on the zone and we forget to update the config ... which one do you prefer? id be ok to add RBAC to read a zone ..

organization-server/internal/client/rover.go
ron96g: However, this will no longer work as we already switched-off the mock-mode on these APIs. Question is what is the ideal solution here. Implement some temporary fix or update common-server for this.

organization-server/internal/client/token.go
ron96g: Why is the wrap needed? Why not use it directly in the http.Client like we do at other places

organization-server/internal/graphql/operations.graphql
```diff
@@ -0,0 +1,199 @@
+# Copyright 2025 Deutsche Telekom IT GmbH
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# GraphQL operations for the organization-server facade.
```
ron96g, line 5: Question: Is there some way to detect drift between this and the actual graphql-schema already at build-time? Otherwise this will first fail at deploy-time, where I can already see that we overlook it and it plays on some customer-facing env :/

organization-server/internal/handler/handler.go
```diff
@@ -0,0 +1,56 @@
...
+		cpapi: cpapi,
+		rover: rover,
+		log:   log,
+	}
+}
+
+// RegisterRoutes registers all REST endpoints on the given Fiber router group.
+// The router group is expected to be mounted at /organization/v1.
+// teamAuth is applied to routes that access hub/team-scoped resources.
+func (h *Handler) RegisterRoutes(api fiber.Router, teamAuth fiber.Handler) {
```
ron96g: Why is this differently implemented than e.g. https://github.com/telekom/controlplane/blob/main/discovery-server/internal/server/server.go or https://github.com/telekom/controlplane/blob/main/rover-server/internal/server/server.go

controlplane-api/internal/service/resource_checker.go
```diff
@@ -0,0 +1,89 @@
...
+
+func (r *roverResourceChecker) HasResources(ctx context.Context, prefix string) (bool, error) {
+	url := fmt.Sprintf("%s/resources?prefix=%s&limit=1", r.baseURL, prefix)
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
+	if err != nil {
+		return false, fmt.Errorf("building request: %w", err)
+	}
+
+	token := mock.NewMockAccessToken(r.environment, "cpapi", "service", []string{r.scopePrefix + ":admin:all"})
```
ron96g: also mock

organization-server/internal/client/rover.go
julius-malcovsky: (waiting for the common-server update)

controlplane-api/internal/service/resource_checker.go
julius-malcovsky: (waiting for the common-server update)

organization-server/internal/client/token.go
julius-malcovsky: good point, simplified

organization-server/internal/handler/handler.go
julius-malcovsky: unified

organization-server/internal/graphql/operations.graphql
julius-malcovsky: (ai answer, but makes sense)

Drift is already detected at build time —  genqlient.yaml  validates  operations.graphql  against the actual CP API schema files ( ../controlplane-api/ent.graphql ,  schema.graphql ,  mutation.graphql ). If a field is renamed/removed,  genqlient  fails during generation.

What was missing: CI wasn't enforcing regeneration. Now:

• Flipped  run_check_generated_files: true  in CI
• CI runs  make generate  +  git diff --exit-code  — catches both schema drift and forgotten regeneration
• Added no-op  manifests  target to Makefile (CI requires it)

Same pattern used by rover-server, discovery-server, and other modules that already have this enabled.

julius-malcovsky: updated the workflows + org-server makefile

iagarba (approved):

ron96g (approved):

PR #486, opened 2026-07-13
feat(controlplane-api,projector): add requested-scopes for approval and approvalrequest

ron96g:


controlplane-api/ent.graphql
```diff
@@ -1289,6 +1293,10 @@ type ApprovalRequest implements Node {
   deciderTeamName: String!
   decisions: [Decision!]!
   availableTransitions: [AvailableTransition!]
+  """
+  If any, the access-scopes requested.
+  """
+  requestedScopes: String
```
stefan-ctrl: So far, we've got scopes desfined as a list: https://github.com/telekom/controlplane/blob/6969e76ce1be9eb401af69a043196bd93f2cbb5c/controlplane-api/ent.graphql#L16
https://github.com/telekom/controlplane/blob/6969e76ce1be9eb401af69a043196bd93f2cbb5c/controlplane-api/schema.graphql#L174

I don't think it makes sense to break this pattern for requestedScopes as an inline comma-separated list.

projector/internal/domain/approval/properties.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+
+	var raw rawProperties
+	if err := json.Unmarshal(requester.Properties.Raw, &raw); err != nil {
+		return props, err
+	}
+
+	props.Scopes = strings.Join(raw.Scopes, ", ")
+
+	// ... extract other fields ...
```
stefan-ctrl, line 37: Seems like there are no other fields :D

projector/internal/domain/approval/properties_test.go
```diff
@@ -0,0 +1,59 @@
...
+		Expect(props.Scopes).To(BeEmpty())
+	})
+
+	It("should join multiple scopes with a comma separator", func() {
+		req := approvalv1.Requester{}
+		Expect(req.SetProperties(map[string]any{"scopes": []string{"read", "write"}})).To(Succeed())
+
+		props, err := approval.FromProperties(req)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(props.Scopes).To(Equal("read, write"))
```
stefan-ctrl: As said, I'd prefer a list here as well, to be aligned with how scopes are handled elsewhere

stefan-ctrl (changes requested):
Seems like a solid change and a nice addition for future UI improvements.

Would prefer scopes to be a list in Approval/ApprovalRequest to be aligned with scopes handling everywhere else.

controlplane-api/ent.graphql
ron96g: Yep, agreed

projector/internal/domain/approval/properties.go
ron96g: But might be in the future ;)

stefan-ctrl (approved):
Thanks for the adjustment. LGTM.

PR #490, opened 2026-07-14
feat(oauth2): add refresh-token 

stefan-ctrl:
add granttype refresh_token for externalIDP from rovers-erver -> jumper

Regarding additional enum checks for refresh_token from https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.3: 
> If the client type is confidential or the client was issued client
   credentials (or assigned other authentication requirements), **the
   client MUST authenticate with the authorization server as described
   in [Section 3.2.1](https://datatracker.ietf.org/doc/html/rfc6749#section-3.2.1).**


gateway/api/v1/security_types.go
```diff
@@ -182,4 +183,8 @@ type OAuth2ClientCredentials struct {
 	// clientKey is the private key associated with the client ID
 	// +kubebuilder:validation:Optional
 	ClientKey string `json:"clientKey,omitempty"`
+	// RefreshToken is an OAuth2 refresh token used to obtain new access tokens
+	// without requiring re-authentication. Used with the refresh_token grant type.
+	// +kubebuilder:validation:Optional
+	RefreshToken string `json:"refreshToken,omitempty"`
```
ron96g, line 189: should there be some further CEL validation similiar to ClientKey?

rover/api/v1/security_types.go
```diff
@@ -121,4 +122,8 @@ type OAuth2ClientCredentials struct {
 	// ClientKey is the private key associated with the client ID
 	// +kubebuilder:validation:Optional
 	ClientKey string `json:"clientKey,omitempty"`
+	// RefreshToken is an OAuth2 refresh token used to obtain new access tokens
+	// without requiring re-authentication. Used with the refresh_token grant type.
+	// +kubebuilder:validation:Optional
+	RefreshToken string `json:"refreshToken,omitempty"`
```
ron96g, line 128: same as in gateway

rover/internal/webhook/v1/secrets.go
```diff
@@ -20,128 +23,87 @@ import (
 	secretsapi "github.com/telekom/controlplane/secret-manager/api"
 )
 
-// TODO: this list of secrets is not complete, it should be extended with all secrets that are used in the rover spec
```
ron96g, line 23: this looks like scope creep ;) What exactly is happening here?

ron96g (changes requested):
Looks good in general. Just some minor stuff regarding CEL so to keep it aligned with the existing things. However, I am not 100% sure about the change in the `secrets.go`. The goal is to make it more generic but what does `strings.Index(path, ".api.")` do? It should be generic in a sense that it can also be used for other resources like event, file etc.

ron96g (approved):

PR #491, opened 2026-07-15
feat: support custom-claims that are added to the lms-token; self-service aud-claims

ron96g:
This PR adds support for custom claims added to the platforms LMS-Token issued by the Gateway-Jumper.
This only applied to the m2m oauth2 config without externalIDP, because:

- basicAuth does not issue a token
- we dont control the token issued by the external IDP
- it can only be applied to upstream m2m-security (User-Application </> Gateway <-> Provider) as it is a provider-side claim

It offers two variants:

- `value`: Is a hardcoded claim-value set in all claims
- `valueFrom`: Is a request-based value-reference which is either (1) resolved by the CP or (2) resolved by the Gateway-Jumper at runtime.

(1) BasePath and ProviderClientId are both known at config-time and does not change/scale with the number of consumers. Its constant for the Exposure
(2) ConsumerClientId is not known at config-time as it depends on the consumer sending the request. It must therefore be resolved at runtime.


julius-malcovsky (dismissed):
nice :)

gateway/internal/features/feature/claims.go
```diff
@@ -0,0 +1,85 @@
...
+func (f *ClaimsFeature) IsUsed(ctx context.Context, builder features.FeaturesBuilder) bool {
+	route, ok := builder.GetRoute()
+	if !ok {
+		return false
+	}
+	notPassThrough := !route.Spec.PassThrough
+	isPrimaryRoute := !route.IsProxy()
+	isFailoverSecondary := route.Spec.Type == gatewayv1.RouteTypeSecondary
+
+	return notPassThrough && (isPrimaryRoute || isFailoverSecondary)
```
iagarba: could you add some comment as when it is used?

docs/docs/user-journey/features/security.mdx
```diff
@@ -94,6 +94,74 @@ exposures:
...
+
+As an API provider, you can require that every token reaching your backend contains a specific audience value. The claim applies to **all** consumers of the exposed API.
+
+**Applies to:** API Exposure
+
+You define the audience under `security.m2m.claims.aud`. You must set **exactly one** of the following:
+
+- `value` — a fixed text value that you choose.
+- `valueFrom` — a predefined source that the platform fills in for you.
+
```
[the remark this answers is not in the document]
luispflamminger: I guess copilot is right here. L92 documents `exposures[].security.scopes` when it should actually be `exposures[].security.m2m.scopes`? Might be better to fix in a separate PR though, as it's unrelated to this change.

gateway/api/v1/security_types.go
```diff
@@ -49,6 +49,28 @@ type Security struct {
...
+// +kubebuilder:validation:Enum=ConsumerClientId
+type ClaimValueFrom string
+
+const (
+	ClaimValueFromConsumerClientId ClaimValueFrom = "ConsumerClientId"
+)
+
+// Claim is a single token claim written into JumperConfig.Claims.
+// Value is a CP-resolved literal; ValueFrom is resolved by Jumper at runtime.
+type Claim struct {
```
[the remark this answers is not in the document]
luispflamminger: This is validated directly in the CRD, so should not be required here.

api/config/crd/bases/api.cp.ei.telekom.de_apisubscriptions.yaml
```diff
@@ -93,6 +93,31 @@ spec:
...
+                                enum:
+                                - ProviderClientId
+                                - ConsumerClientId
+                                - BasePath
+                                type: string
+                            type: object
+                            x-kubernetes-validations:
+                            - message: exactly one of value or valueFrom must be set
+                              rule: has(self.value) != has(self.valueFrom)
+                        type: object
```
luispflamminger: Do we already want to implement the consumer override now? I thought it was for a possible future extension?

gateway/internal/features/feature/claims.go
ron96g: Personally I think this code already speaks for itself, or what do you mean exactly?

- Its not a passthrough route
- Its either a primary or a failoverSecondary Route

iagarba: Yes it does speak for itself and I understand it, just had to lookup "failoverSecondary" again...

rover/api/v1/rover_types.go
```diff
@@ -145,13 +145,16 @@ type RoverSpec struct {
 
 	// Exposures is a list of APIs and Events that this Rover exposes to consumers
 	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:MaxItems=50
```
luispflamminger, line 148: Why are these added here?

luispflamminger (commented):
Might be some dumb questions, I'm new to this 🙃

api/config/crd/bases/api.cp.ei.telekom.de_apisubscriptions.yaml
ron96g: You are right. It was just really easy to implement ;) But in practice its not used at the moment and we also dont have any concrete plans. I would remove it for now, if we have a use-case for it, we can always add it back. No harm done

rover/api/v1/rover_types.go
ron96g: k8s CEL Validation rules have a cost-factor attached to them. So the more CEL validation (especially complex XValidation-rules) you add, the more costly they become and are at some point refused by k8s - this happened here. A solution to this is, to add MaxLength and MaxItems markers. 
So it is kind of a scope-creep but it solved an immediate issue and should not cause harm as we dont want these super-large Rovers anyways.

docs/docs/user-journey/features/security.mdx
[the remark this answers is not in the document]
ron96g: Yep, it probably mixed the internal format with the user-facing format... Will check

rover/api/v1/rover_types.go
luispflamminger: Interesting! Thanks for the explanation.

docs/docs/user-journey/features/security.mdx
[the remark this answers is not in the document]
ron96g: fixed

api/config/crd/bases/api.cp.ei.telekom.de_apisubscriptions.yaml
ron96g: removed the consumer side

luispflamminger: great, thx!

luispflamminger (dismissed):

iagarba (approved):

PR #500, opened 2026-07-21
fix: resource_type is set by notification-sender

ron96g:


julius-malcovsky (approved):
LGTM - thanks for fixing it

PR #507, opened 2026-07-24
feat(servers): enhance server-security by adding listeners for internal and external traffic

ron96g:
This PR adds support for multiport listeners enabling us to publish a server for internal and external use on different ports.
In addition both listener may be configured with different auth thus enabling internal-listeners to use k8s-jwt (in-cluster) and external-listener to use normal jwt-lms (keycloak).

Additional changes:
- feat: aligned all config so that each components just embed the BasisConfig from common-server via composition.
- feat(rover-server): now listening via TLS on both listeners. This was http before and needed to be changed anyways.


common-server/pkg/server/listener_config.go
```diff
@@ -0,0 +1,153 @@
...
+func (c *TLSFileConfig) ToServerTLS() *TLSConfig {
+	if c == nil {
+		return nil
+	}
+	return &TLSConfig{CertFile: c.Cert, KeyFile: c.Key}
+}
+
+// K8sConfig is the declarative config for the Kubernetes-authz security family.
+type K8sConfig struct {
+	Audience       string                    `mapstructure:"audience"`
```
ron96g, line 32: Make this an array

common-server/pkg/server/middleware/security/synthetic_context.go
```diff
@@ -0,0 +1,57 @@
...
+//
+// Behaviour is fixed by the settled design and takes no options:
+//   - Environment comes from the mandatory X-Environment header.
+//   - ClientType is always admin; AccessType is always read-write ("all").
+//   - The store prefix is "<env>--" (the admin template).
+//
+// A missing or empty X-Environment is rejected with 400 — a blank env would
+// yield the prefix "--", which store filtering treats as *all* environments.
+func NewSyntheticAdminBusinessContext() fiber.Handler {
+	return func(c *fiber.Ctx) error {
```
ron96g, line 30: where is the sub-claim to set the name

common-server/pkg/server/middleware/security/synthetic_context.go
```diff
@@ -0,0 +1,57 @@
...
+			p := problems.BadRequest("Missing or empty X-Environment header")
+			return c.Status(p.Code()).JSON(p, "application/problem+json")
+		}
+
+		bCtx := &BusinessContext{
+			Environment: env,
+			ClientType:  ClientTypeAdmin,
+			AccessType:  AccessTypeReadWrite,
+		}
+		prefix := env + "--"
```
ron96g, line 42: "--" is this correct? Can it be a different prefix

common-server/pkg/server/middleware/security/synthetic_context.go
```diff
@@ -0,0 +1,57 @@
...
+// Behaviour is fixed by the settled design and takes no options:
+//   - Environment comes from the mandatory X-Environment header.
+//   - ClientType is always admin; AccessType is always read-write ("all").
+//   - The store prefix is "<env>--" (the admin template).
+//
+// A missing or empty X-Environment is rejected with 400 — a blank env would
+// yield the prefix "--", which store filtering treats as *all* environments.
+func NewSyntheticAdminBusinessContext() fiber.Handler {
+	return func(c *fiber.Ctx) error {
+		env := c.Get("X-Environment")
```
ron96g, line 31: idea: Add some X-User-Identity header to assume a team-user

common-server/pkg/server/middleware/security/synthetic_context.go
ron96g: verified, is fine

common-server/pkg/server/middleware/security/synthetic_context.go
ron96g: YAGNI

common-server/pkg/server/middleware/security/synthetic_context.go
ron96g: YAGNI

julius-malcovsky (approved):
nicely done

PR #512, opened 2026-07-30
fix(api): clean up stale ApiExposure routes on zone change

ron96g:
Moving an ApiExposure's zone changes the derived real-route namespace, so
CreateOrUpdate created a new real route and orphaned the old one. Real routes
were excluded from the proxy-route cleanup, leaving no path to remove them on a
normal reconcile.

Unify route cleanup on a single basepath-scoped janitor sweep
(CleanupStaleRoutes) run after all routes are provisioned, covering proxy,
failover, and real routes. Align ApiMustExist to the same sweep so all routes
are removed when the Api is deregistered. Stamp an informational owner.uid
label on routes for traceability.


julius-malcovsky (approved):
should work :)

PR #531, opened 2026-08-03
feat(file): implement file domain

lukas016:
This PR introduces the new file domain to controlplane by adding the v1 API types and generated CRD manifests for FileType, FileExposure, FileSubscription, and ZoneServiceConfig, along with the required kubebuilder registration, typed references, label utilities, and validation/webhook scaffolding needed to manage file-based resources and zone-scoped configuration in the platform. It also includes test coverage for handler and webhook behavior to confirm the expected lifecycle and validation logic.


file/api/v1/shared_types.go
```diff
@@ -0,0 +1,75 @@
...
+	VisibilityZone       Visibility = "Zone"
+	VisibilityEnterprise Visibility = "Enterprise"
+)
+
+// ApprovalStrategy defines the approval mode for subscriptions.
+// +kubebuilder:validation:Enum=Auto;Simple;FourEyes
+type ApprovalStrategy string
+
+const (
+	ApprovalStrategyAuto     ApprovalStrategy = "Auto"
```
lukas016, line 44: check if this can be imported

file/api/go.sum.license
```diff
@@ -0,0 +1,3 @@
+Copyright 2025 Deutsche Telekom IT GmbH
```
lukas016: this should be 2026

file/config/default/kustomization.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+labels:
+- includeSelectors: true
+  pairs:
+    domain: file
+
+resources:
+- ../crd
+- ../rbac
+- ../manager
```
lukas016, line 28: missing webhook

file/internal/handler/filesubscription/handler.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+// teamNameFromNamespace extracts the composite team name from a namespace
+// following the convention "<env>--<group>--<team>".
+//
+// The Team CR metadata.name is "<group>--<team>" (e.g. "eni--narvi-regr"),
+// so we drop the first segment (environment) and return everything after
+// the first "--" separator. If no "--" is found the full namespace is
+// returned as-is.
+func teamNameFromNamespace(namespace string) string {
```
lukas016, line 209: this can be part of common. add todo comment

file/internal/handler/util/getters.go
```diff
@@ -0,0 +1,110 @@
...
+			return nil, ctrlerrors.BlockedErrorf("FileType %q not found", ref.String())
+		}
+		return nil, fmt.Errorf("failed to get FileType %q: %w", ref.String(), err)
+	}
+	return fileType, nil
+}
+
+func GetChildResourceRef(obj *filev1.ZoneServiceConfig) types.ObjectRef {
+	return types.ObjectRef{
+		Name:      identityClientNamePrefix + "--" + obj.Name,
```
lukas016, line 42: @ron96g  do we have some constant for --  in the code?

file/internal/handler/util/publickeys.go
```diff
@@ -0,0 +1,131 @@
...
+	if err := c.Delete(ctx, instance); err != nil {
+		if apierrors.IsNotFound(errors.Cause(err)) {
+			return nil
+		}
+		return fmt.Errorf("failed to delete SFTP Instance %q: %w", instanceRef.String(), err)
+	}
+	return nil
+}
+
+func CanonicalSSHPublicKeys(publicKeys []filev1.SSHPublicKeySpec) ([]string, error) {
```
lukas016, line 98: whole function can be simplified

file/internal/handler/zoneserviceconfig/handler.go
```diff
@@ -0,0 +1,360 @@
...
+
+	routeConfig := &obj.Spec.API
+
+	mutator := func() error {
+		locErr := controllerutil.SetControllerReference(obj, route, cc.Scheme())
+		if locErr != nil {
+			return fmt.Errorf("failed to set controller reference for route: %w", locErr)
+		}
+
+		if route.Labels == nil {
```
lukas016: leftover

file/api/v1/shared_types.go
lukas016: every operator has own copy. Keep it as it is for now

file/internal/controller/fileexposure_controller.go
```diff
@@ -0,0 +1,116 @@
...
+
+func (r *FileExposureReconciler) MapFileTypeToFileExposure(ctx context.Context, obj client.Object) []reconcile.Request {
+	fileType, ok := obj.(*filev1.FileType)
+	if !ok {
+		return nil
+	}
+
+	list := &filev1.FileExposureList{}
+	err := r.List(ctx, list,
+		client.InNamespace(fileType.Namespace),
```
ron96g: The FileType can be in a different NS. e.g. specified by team-a and exposed by team-b

file/internal/controller/fileexposure_controller.go
```diff
@@ -0,0 +1,116 @@
...
+
+func (r *FileExposureReconciler) MapZoneServiceConfigToFileExposure(ctx context.Context, obj client.Object) []reconcile.Request {
+	zoneServiceConfig, ok := obj.(*filev1.ZoneServiceConfig)
+	if !ok {
+		return nil
+	}
+
+	list := &filev1.FileExposureList{}
+	if err := r.List(ctx, list,
+		client.InNamespace(zoneServiceConfig.Labels[cconfig.EnvironmentLabelKey]),
```
ron96g: The FileExposure is in the team-NS or not? So it cannot be selected via environment

file/api/v1/fileexposure_types.go
```diff
@@ -0,0 +1,99 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/telekom/controlplane/common/pkg/types"
+)
+
+// FileExposureSpec defines a provider-side file exposure.
+type FileExposureSpec struct {
+	// Provider optionally identifies the providing application.
+	// +kubebuilder:validation:Optional
+	Provider string `json:"provider,omitempty"`
```
ron96g: Why optionally? And why string in favour of ObjectRef?

file/api/v1/fileexposure_types.go
```diff
@@ -0,0 +1,99 @@
...
+
+	// +kubebuilder:validation:Required
+	FileType string `json:"fileType"`
```
ron96g: would be nice to have a comment on here. Like "This is the unique identifier of the FileType that is being exposed"

file/api/v1/shared_types.go
```diff
@@ -0,0 +1,77 @@
...
+
+// FileSolution defines the backing file transfer solution.
+// +kubebuilder:validation:Enum=sftp
+type FileSolution string
+
+const (
+	FileSolutionSFTP FileSolution = "sftp"
+)
+
+func (s FileSolution) OrDefault() FileSolution {
```
ron96g: dead code?

file/api/go.mod
```diff
@@ -0,0 +1,59 @@
...
+	k8s.io/kube-openapi v0.0.0-20260317180543-43fb72c5454a // indirect
+	k8s.io/utils v0.0.0-20260210185600-b8788abfbbc2 // indirect
+	sigs.k8s.io/json v0.0.0-20250730193827-2d320260d730 // indirect
+	sigs.k8s.io/randfill v1.0.0 // indirect
+	sigs.k8s.io/structured-merge-diff/v6 v6.3.2 // indirect
+	sigs.k8s.io/yaml v1.6.0 // indirect
+)
+
+replace (
+	github.com/telekom/controlplane/admin/api => ../../admin/api
```
ron96g: in the api-module we only allow imports of common and not other domains

file/api/v1/zoneserviceconfig_types.go
```diff
@@ -0,0 +1,109 @@
...
+	// +kubebuilder:validation:Optional
+	ServiceExternal *ServiceEndpoint `json:"serviceExternal"`
+
+	// Zone identifies the zone where this file exposure is provided.
+	// +kubebuilder:validation:Required
+	Zone *types.ObjectRef `json:"zone"`
+}
+
+// ServiceEndpoint identifies an SFTP service endpoint.
+type ServiceEndpoint struct {
```
ron96g: This is always https, right?

file/api/v1/zoneserviceconfig_types.go
```diff
@@ -0,0 +1,109 @@
...
+
+	adminv1 "github.com/telekom/controlplane/admin/api/v1"
+	"github.com/telekom/controlplane/common/pkg/types"
+)
+
+// ZoneServiceConfigSpec defines the file-domain service configuration for one zone.
+// A ZoneServiceConfig must use the same name and namespace as its admin Zone.
+type ZoneServiceConfigSpec struct {
+	// +kubebuilder:validation:Required
+	API adminv1.ManagedRouteConfig `json:"api"`
```
ron96g: This is not needed, right? its never a TeamAPI, it would always be an internal-route

file/api/v1/zoneserviceconfig_types.go
```diff
@@ -0,0 +1,109 @@
...
+
+	// Service is the internal SFTP service endpoint.
+	// +kubebuilder:validation:Optional
+	Service *ServiceEndpoint `json:"service"`
```
ron96g: Why is this needed? Cant we get this info from the referenced Zone?

file/api/v1/zoneserviceconfig_types.go
```diff
@@ -0,0 +1,109 @@
...
+
+	// ServiceExternal is the externally reachable SFTP service endpoint.
+	// +kubebuilder:validation:Optional
+	ServiceExternal *ServiceEndpoint `json:"serviceExternal"`
```
ron96g: Why is this optional, we always need an external service or not?

file/internal/handler/fileexposure/handler.go
```diff
@@ -0,0 +1,140 @@
...
+	fileTypeRef := types.ObjectRef{Namespace: obj.Namespace, Name: obj.Spec.FileType}
+
+	activeExposure, found, err := util.FindActiveFileExposure(ctx, &fileTypeRef)
+	if err != nil {
+		return err
+	}
+
+	if found && activeExposure.UID != obj.UID {
+		obj.Status.FileTypeRef = &fileTypeRef
+		obj.SetCondition(condition.NewNotReadyCondition("FileExposureAlreadyExists", "Another FileExposure already provides this FileType"))
```
ron96g: check the common conditions and change it if appliable, e.g. ReasonPreconditionNotMet

file/internal/handler/filesubscription/handler.go
```diff
@@ -0,0 +1,214 @@
...
+
+var _ handler.Handler[*filev1.FileSubscription] = &FileSubscriptionHandler{}
+
+type FileSubscriptionHandler struct{}
+
+func (h *FileSubscriptionHandler) CreateOrUpdate(ctx context.Context, obj *filev1.FileSubscription) error {
+	logger := log.FromContext(ctx)
+	c := cclient.ClientFromContextOrDie(ctx)
+
+	fileType, err := util.GetFileType(ctx, types.ObjectRef{Namespace: obj.Namespace, Name: obj.Spec.FileType})
```
ron96g: Your assumption is that the FileType is always specified by the same team that is subscribing. I think that this cannot be assumed. You must List and selected via labels and active-flag.

file/internal/handler/filesubscription/handler.go
```diff
@@ -0,0 +1,214 @@
...
+	res, err := h.ensureApproval(ctx, obj, fileType, activeExposure)
+	if err != nil {
+		return err
+	}
+	switch res {
+	case builder.ApprovalResultRequestDenied:
+		logger.Info("ApprovalRequest was denied - deleting subscriber SFTP User")
+		obj.SetCondition(condition.NewNotReadyCondition("ApprovalRequestDenied", "ApprovalRequest has been denied"))
+		obj.SetCondition(condition.NewDoneProcessingCondition("ApprovalRequest has been denied"))
+		return h.deleteSubscriberUser(ctx, obj)
```
ron96g: This should not be deleted, right? If only the Request was denied but there is an active Approval (granted), then this should block but not delete

file/internal/handler/filesubscription/handler.go
```diff
@@ -0,0 +1,214 @@
...
+
+	err := requester.SetProperties(properties)
+	if err != nil {
+		return builder.ApprovalResultNone, fmt.Errorf("unable to set approvalRequest properties for FileSubscription %q in namespace %q: %w",
+			obj.Name, obj.Namespace, err)
+	}
+
+	decider := &approvalapi.Decider{
+		TeamName:       teamNameFromNamespace(activeExposure.Namespace),
+		ApplicationRef: types.TypedObjectRefFromObject(activeExposure, c.Scheme()),
```
ron96g: This looks wrong. This returns the Ref to the FileExposure but not the ApplicationRef, right?

file/internal/handler/filesubscription/handler.go
```diff
@@ -0,0 +1,214 @@
...
+		return builder.ApprovalResultNone, err
+	}
+	obj.Status.ApprovalRequest = types.ObjectRefFromObject(approvalBuilder.GetApprovalRequest())
+	obj.Status.Approval = types.ObjectRefFromObject(approvalBuilder.GetApproval())
+
+	return res, nil
+}
+
+func (h *FileSubscriptionHandler) syncSubscriberUser(ctx context.Context, obj *filev1.FileSubscription, fileType *filev1.FileType, activeExposure *filev1.FileExposure) error {
+	_, err := util.SyncSFTPUser(
```
ron96g: [nitpick] maybe refactor to reduce complexity? A lot of params here

file/internal/handler/filesubscription/handler.go
```diff
@@ -0,0 +1,214 @@
...
+
+func approvalProperties(subscription *filev1.FileSubscription) map[string]any {
+	return map[string]any{
+		"fileType": subscription.Spec.FileType,
+		"zone":     subscriptionZoneName(subscription),
+	}
+}
+
+func subscriptionZoneName(subscription *filev1.FileSubscription) string {
+	if subscription.Spec.Zone == nil {
```
ron96g: This must never happen. It must be a  required field

file/internal/handler/filetype/handler.go
```diff
@@ -0,0 +1,60 @@
...
+		return err
+	}
+
+	obj.Status.FileExposureRef = types.ObjectRefFromObject(activeExposure)
+	obj.Status.SFTPInstance = &types.ObjectRef{
+		Name:      obj.Name,
+		Namespace: obj.Namespace,
+	}
+
+	if !c.AllReady() {
```
ron96g: Is this creating anything or dead-code?

file/internal/handler/filetype/handler.go
```diff
@@ -0,0 +1,60 @@
...
+		return nil
+	}
+
+	_, err = util.GetZoneServiceConfig(ctx, activeExposure.Spec.Zone)
+	if err != nil {
+		return err
+	}
+
+	obj.Status.FileExposureRef = types.ObjectRefFromObject(activeExposure)
+	obj.Status.SFTPInstance = &types.ObjectRef{
```
ron96g: dead code?

file/internal/handler/util/getters.go
ron96g: Sorry for the late reply: No, but its also not really needed here. The "--" is only used to split customer-provided values

file/internal/handler/util/getters.go
```diff
@@ -0,0 +1,110 @@
...
+
+	zoneServiceConfig := &list.Items[0]
+	return zoneServiceConfig, nil
+}
+
+func FindFileExposuresForFileType(ctx context.Context, fileType *types.ObjectRef) ([]filev1.FileExposure, error) {
+	c := cclient.ClientFromContextOrDie(ctx)
+	list := &filev1.FileExposureList{}
+	if err := c.List(ctx, list,
+		client.InNamespace(fileType.Namespace),
```
ron96g: It can be accross namespaces, e.g. different teams expose the same FileType

file/internal/handler/util/getters.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func GetChildResourceRef(obj *filev1.ZoneServiceConfig) types.ObjectRef {
+	return types.ObjectRef{
+		Name:      identityClientNamePrefix + "--" + obj.Name,
+		Namespace: obj.Namespace,
+	}
+}
+
+func GetZoneServiceConfig(ctx context.Context, ref *types.ObjectRef) (*filev1.ZoneServiceConfig, error) {
```
ron96g: What is ref here?

file/internal/handler/util/getters.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func GetPublicKeysFromSFTP(sftp *filev1.FileSFTP) []filev1.SSHPublicKeySpec {
+	if sftp == nil {
+		return nil
+	}
+	return sftp.PublicKeys
+}
+
+func GetZoneNamespace(ref *types.ObjectRef) string {
```
ron96g: I dont like this convention. Why not fetch Zone, get Zone.Status.Namespace and use it explictly. Yes, its an additional cache-lookup for Zone but it makes it clearer in my opinion.

file/internal/handler/util/labels.go
```diff
@@ -0,0 +1,31 @@
...
+		config.DomainLabelKey:                 "file",
+		filev1.FileTypeNameLabelKey:           labelutil.NormalizeLabelValue(fileTypeRef.Name),
+		filev1.FileTypeNamespaceLabelKey:      labelutil.NormalizeLabelValue(fileTypeRef.Namespace),
+		config.BuildLabelKey("file.exposure"): labelutil.NormalizeLabelValue(fileTypeRef.Name),
+		config.BuildLabelKey("managed.by"):    "file-operator",
+	}
+}
+
+// FileTypeLabelSelector returns a selector matching resources labeled for the given FileType name.
+func FileTypeLabelSelector(fileTypeName string) labels.Selector {
```
ron96g: dead code?

file/api/v1/filetype_types.go
```diff
@@ -0,0 +1,83 @@
...
+
+// FileTypeSpec defines a logical file type that can be exposed and subscribed to.
+type FileTypeSpec struct {
+	// Description is a human-readable description of this file type.
+	// +kubebuilder:validation:Optional
+	Description string `json:"description,omitempty"`
+}
+
+// FileTypeStatus defines the observed state of FileType.
+type FileTypeStatus struct {
```
ron96g: How do you identify the active FileType? Or are all active but only one FileExposure? This break the current pattern

file/cmd/file-operator/main.go
```diff
@@ -0,0 +1,185 @@
...
+	if metricsCertPath != "" {
+		setupLog.Info("Initializing metrics certificate watcher using provided certificates",
+			"metrics-cert-path", metricsCertPath, "metrics-cert-name", metricsCertName, "metrics-cert-key", metricsCertKey)
+
+		metricsServerOptions.CertDir = metricsCertPath
+		metricsServerOptions.CertName = metricsCertName
+		metricsServerOptions.KeyName = metricsCertKey
+	}
+
+	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
```
ron96g: I propose to add a cache-selector similiar to event to avoid caching Consumer, Routes, etc. that do not belong to this domains as otherwise this will drastically increase the cache-size without any benefit

file/internal/handler/zoneserviceconfig/handler.go
```diff
@@ -0,0 +1,356 @@
...
+	if found {
+		if !condition.IsReady(apiClient) {
+			return apiClient, nil
+		}
+
+		if apiClient.Status.SecretExpiresAt == nil {
+			return apiClient, nil
+		}
+
+		weekBeforeExp := apiClient.Status.SecretExpiresAt.Add(-7 * 24 * time.Hour)
```
ron96g: Internal realms will have the secret-rotation disabled for now. Thus, this is dead-code. I would propose to remove it and then in the future when it comes to it, we need to find a central solution

file/internal/handler/zoneserviceconfig/handler.go
```diff
@@ -0,0 +1,356 @@
...
+		if time.Now().Before(weekBeforeExp) {
+			return apiClient, nil
+		}
+	}
+
+	clientSecretPath := fmt.Sprintf("zones/%s/file/%s/%s/clientSecret", zone.Name, obj.Namespace, obj.Name)
```
ron96g: Why is namespace and name needed? The idea is 1 ZoneConfig per Zone

file/internal/handler/zoneserviceconfig/handler.go
```diff
@@ -0,0 +1,356 @@
...
+		if apierrors.IsNotFound(errors.Cause(err)) {
+			return nil, ctrlerrors.BlockedErrorf("Zone %q not found", ref.String())
+		}
+		return nil, fmt.Errorf("failed to get Zone %q: %w", ref.String(), err)
+	}
+	return zone, nil
+}
+
+// createOrUpdateSFTPAPIClient creates or updates an identity Client for the SFTP API.
+func createOrUpdateSFTPAPIClient(ctx context.Context, obj *filev1.ZoneServiceConfig, zone *adminv1.Zone) (*identityv1.Client, error) {
```
ron96g: I would propose to refactor this function and remove the creation part into a separate function?

file/internal/handler/zoneserviceconfig/handler.go
```diff
@@ -0,0 +1,356 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate secret for SFTP API Client: %w", err)
+	}
+
+	options := []secretsapi.OnboardingOption{
+		secretsapi.WithMergeStrategy(),
+		secretsapi.WithSecretValue(clientSecretPath, secretValue),
+	}
+
+	availableSecret, err := secretsapi.API().UpsertEnvironment(ctx, zone.Labels[cconfig.EnvironmentLabelKey], options...)
```
ron96g: zone.Labels[cconfig.EnvironmentLabelKey] can be replaced by contextutil.EnvFromContextOrDie

ron96g (changes requested):
This review was done by hand ;) I might ask AI to review it also, this might result in more comments:
- Please check the conditions in common and align where applicable
- Think about the caching improvement in main.go
- I think the idea with the Active-Exposure via FileType-Status is fine but it does break our current pattern

file/internal/controller/fileexposure_controller.go
lukas016: we have open discussion about it somewhere, if we want to support it. Because you can still data with this approach from another team.

file/internal/handler/zoneserviceconfig/handler.go
lukas016: yes it is. but this for me more user-friendly because you have direct mapping between zoneserviceconfig and related credential

file/internal/handler/zoneserviceconfig/handler.go
lukas016: fixed

file/internal/handler/zoneserviceconfig/handler.go
lukas016: do you need central solution for it? Technically this client is created and manage by sftp operator. SO it should control secret rotation too.

file/internal/handler/util/getters.go
lukas016: i don't think it is true. I saw split namespace with --: eni--galatea. And i don't think it is customer provided value.

file/api/v1/shared_types.go
lukas016: fixed

file/internal/controller/fileexposure_controller.go
lukas016: do it as before

file/internal/handler/zoneserviceconfig/handler.go
lukas016: without resource name

file/internal/controller/fileexposure_controller.go
lukas016: fixed

file/internal/controller/fileexposure_controller.go
lukas016: fixed

file/api/v1/fileexposure_types.go
lukas016: we will implement it later as additional field.

file/api/v1/fileexposure_types.go
lukas016: fixed

file/internal/handler/zoneserviceconfig/handler.go
lukas016: fixed

file/internal/handler/zoneserviceconfig/handler.go
lukas016: fixed

file/internal/handler/zoneserviceconfig/handler.go
lukas016: fixed

file/cmd/file-operator/main.go
lukas016: fixed

file/internal/handler/zoneserviceconfig/handler.go
lukas016: pushed

file/internal/handler/zoneserviceconfig/handler.go
lukas016: pushed

file/internal/handler/zoneserviceconfig/handler.go
lukas016: pushed

file/internal/handler/zoneserviceconfig/handler.go
lukas016: pushed

file/cmd/file-operator/main.go
lukas016: pushed

file/api/v1/shared_types.go
lukas016: pushed

file/api/v1/fileexposure_types.go
lukas016: pushed

file/internal/controller/fileexposure_controller.go
lukas016: pushed

file/internal/controller/fileexposure_controller.go
lukas016: pushed

file/internal/handler/util/getters.go
lukas016: fixed && pushed

file/api/v1/zoneserviceconfig_types.go
lukas016: this is needed for MC. Question is if you want to have it in zone because file domain is option feature.

file/api/v1/zoneserviceconfig_types.go
lukas016: not always. Minimally i don't know it is supported on playground on all zones. SO i did it more flexible. You can choose if zone has internal or external or both.

file/internal/handler/filesubscription/handler.go
lukas016: this was fixed, i am using list and index

file/internal/handler/util/getters.go
lukas016: this was fixed

file/internal/handler/util/labels.go
lukas016: fixed

file/internal/handler/util/getters.go
lukas016: ok why not

file/internal/handler/util/getters.go
lukas016: fixed

file/internal/handler/filesubscription/handler.go
lukas016: fixed

file/internal/handler/filesubscription/handler.go
lukas016: fixed

file/internal/handler/filesubscription/handler.go
lukas016: fixed

file/api/v1/zoneserviceconfig_types.go
lukas016: no, it should be sftp protocol

file/internal/handler/fileexposure/handler.go
lukas016: fixed

file/api/v1/zoneserviceconfig_types.go
lukas016: add omitempty

file/api/go.mod
lukas016: fixed

file/api/v1/zoneserviceconfig_types.go
lukas016: fixed

file/internal/handler/filetype/handler.go
lukas016: fixed

file/internal/handler/filetype/handler.go
lukas016: this should be correct now.

file/internal/handler/filesubscription/handler.go
lukas016: later

PR #533, opened 2026-08-03
fix(controllers): prevent deleting of referenced realms and gateways

ron96g:


gateway/internal/handler/gateway/handler.go
```diff
@@ -23,7 +28,27 @@ func (h *GatewayHandler) CreateOrUpdate(ctx context.Context, gw *gatewayv1.Gatew
 	return nil
 }
 
-func (h *GatewayHandler) Delete(ctx context.Context, object *gatewayv1.Gateway) error {
+func (h *GatewayHandler) Delete(ctx context.Context, gw *gatewayv1.Gateway) error {
+	kubeClient := cc.ClientFromContextOrDie(ctx)
+	routes := &gatewayv1.RouteList{}
+	if err := kubeClient.List(ctx, routes,
```
julius-malcovsky, line 34: wondering if it would make sense to add the client.Limit(1) for the listing .. were curious if there are any routes at all, we dont need all of them ... yes this is not a hotpath so its nitpicking :) ... plus we would lose the nice error message ... just an idea :)

gateway/internal/controller/gateway_controller.go
```diff
@@ -44,9 +47,31 @@ func (r *GatewayReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&v1.Gateway{}).
+		Watches(&v1.Route{},
```
julius-malcovsky, line 50: can you remind me whats the point of reprocessing the gateway everytime a route is deleted ? i know we talked about it, just cant recall

gateway/internal/handler/gateway/handler.go
ron96g: Yes or/and only get metadata if possible

gateway/internal/controller/gateway_controller.go
ron96g: For correctness to cleanup but I think we should use a RetryWithDelay error instead of a Watcher... Also the Route should watch the Gateway (Route -> gen_change -> Gateway)

stefan-ctrl:
Talked in the team; will leave this PR open for now and discuss it in more detail after the German vacation time.

PR #534, opened 2026-08-03
fix(api): apiexposure must watch application to avoid not-ready-block

ron96g:


BjoernKarma (approved):
LGTM - as discussed in team review

PR #536, opened 2026-08-04
feat(rover): support for pagination

ron96g:
- **fix: backend-protocol must be https**
- **feat(rover-server): add paginated team resource inventory**


controlplane-api/internal/service/resource_checker.go
```diff
@@ -61,17 +73,27 @@ func (r *roverResourceChecker) HasResources(ctx context.Context, group, team str
...
+	defer func() {
+		if closeErr := resp.Body.Close(); err == nil && closeErr != nil {
+			err = fmt.Errorf("closing rover-server response: %w", closeErr)
+		}
+	}()
+	body, err := io.ReadAll(io.LimitReader(resp.Body, maxRoverResponseBytes+1))
+	if err != nil {
+		return false, fmt.Errorf("reading rover-server response: %w", err)
+	}
+	if len(body) > maxRoverResponseBytes {
```
julius-malcovsky, line 88: is this actually necessary ? were putting a limit 1 there - this really is a check if there is anything at all .. do you think that response limited to 1 would exceed 1MB ? .. unless were lucky and we hit that 150 exposures rover ...

julius-malcovsky (dismissed):
just a minor question .. otherwise ok, interesting solution :)

controlplane-api/internal/service/resource_checker.go
ron96g: This is just overengineering by AI I guess. I left it in there as it does not harm us and we set the limit=1 anyways here so it will never be > 1MB

julius-malcovsky (approved):

PR #440, opened 2026-06-26
feat(sftp): implement sftp domain

lukas016:
- sftp domain represent runtime domain
- sftp domain implement management of ssh users and public keys for external sftp service

PR #537, opened 2026-08-04
feat(sftp): implement sftp domain (#440)

lukas016:


sftp/api/v1/instance_types.go
```diff
@@ -0,0 +1,96 @@
...
+// InstanceUserStatus contains the Instance-observed status for a User.
+type InstanceUserStatus struct {
+	// Namespace is the namespace of the User.
+	Namespace string `json:"namespace"`
+
+	// Name is the name of the User.
+	Name string `json:"name"`
+
+	// ProcessingCondition is the User Processing condition observed by the Instance reconciliation.
+	ProcessingCondition metav1.Condition `json:"processingCondition"`
```
ron96g: why not keep jus the LastObservedGeneration isntead of the whiole condition?

sftp/config/manager/manager.yaml
```diff
@@ -0,0 +1,66 @@
...
+            path: /readyz
+            port: 8081
+          initialDelaySeconds: 5
+          periodSeconds: 10
+        resources:
+          limits:
+            cpu: 500m
+            memory: 128Mi
+          requests:
+            cpu: 10m
```
ron96g: must be 30m. Check other operators. Also check if securityContext matches others

sftp/internal/service/http_service.go
```diff
@@ -0,0 +1,98 @@
...
+	client, err := newClientWithResponses(cfg)
+	if err != nil {
+		return nil, err
+	}
+	return &HTTPService{client: client}, nil
+}
+
+func newClientWithResponses(cfg Config) (ClientWithResponsesInterface, error) {
+	httpClient := cfg.HTTPClient
+	if httpClient == nil {
```
ron96g: why not client from common-server? To include metrics, etc.

ron96g (commented):
Looks good. Just a few things that might be blocking. Other than that - if the tests are successful - we can merge it

sftp/config/manager/manager.yaml
lukas016: fixed

sftp/api/v1/instance_types.go
lukas016: this is leftover. It isn't used. i will remove it.

sftp/internal/service/http_service.go
lukas016: fixed plus this condition is deleted. It cannot be nil here.

PR #538, opened 2026-08-04
perf(controller): skip noop status writes

ron96g:
- **chore(pre-commit): restrict reuse-lint-file to the pre-commit stage**
- **fix(api,event,agentic): sort List-derived collections before storing them**
- **perf(controller): skip status updates that change nothing**


common/pkg/test/deepcopy_test .go
```diff
@@ -4,17 +4,24 @@
 
 package test
 
-import "k8s.io/apimachinery/pkg/runtime"
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
```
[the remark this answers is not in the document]
julius-malcovsky: seems to be an empty space at the end of the file name :)

julius-malcovsky (approved):
just fix the filename typo :) gj

julius-malcovsky (dismissed):

api/internal/handler/util/getters.go
```diff
@@ -292,6 +292,12 @@ func FindAllSubscribersForApiExposure(ctx context.Context, apiExp *apiv1.ApiExpo
 		subscribers = append(subscribers, sub)
 	}
 
+	// Cache List order is unstable; this feeds status, so an unsorted result
+	// rewrites the status every reconcile and re-triggers the watch.
+	slices.SortFunc(subscribers, func(a, b *apiv1.ApiSubscription) int {
+		return strings.Compare(a.Namespace+"/"+a.Name, b.Namespace+"/"+b.Name)
```
lukas016: this concat operation isn't necessary and it can do extra allocation/operations. I recommend:

```golang
namespace := strings.Compare(a.Namespace, b.Namespace)
if namespace == 0 {
  return string.Compare(a.Name, b.Name)
}
return namespace
```

common/pkg/controller/controller.go
```diff
@@ -133,6 +134,41 @@ func (c *ControllerImpl[T]) Reconcile(ctx context.Context, req reconcile.Request
...
+	recordStatusUpdate(statusUpdateResultUpdated, object, c.Scheme)
+	return nil
+}
+
+// statusValue returns the object's Status field. Objects without a Status field
+// report false, so the caller falls back to writing rather than comparing.
+func statusValue(object any) (any, bool) {
+	field := reflect.ValueOf(object).Elem().FieldByName("Status")
+	if !field.IsValid() {
+		return nil, false
```
lukas016, line 168: FYI: this negation operation isn't necessary. You can just switch return and logic will act the same without extra operation. Real impact on the performance is almost zero. So it is about personal preferences.

common/pkg/controller/metric.go
```diff
@@ -0,0 +1,41 @@
...
+	Help:      "Total number of controller status update outcomes.",
+}, []string{"result", "kind"})
+
+func init() {
+	metrics.Registry.MustRegister(statusUpdatesTotal)
+}
+
+func recordStatusUpdate(result statusUpdateResult, object runtime.Object, scheme *runtime.Scheme) {
+	kind := "unknown"
+	if gvk, err := apiutil.GVKForObject(object, scheme); err == nil {
```
lukas016, line 37: i think log should be logged as warning (i know warning isn't supported) or debug level.
Because this shouldn't happen in most cases and silent error won't help us find root cause of bug.

common/pkg/controller/metric.go
```diff
@@ -0,0 +1,41 @@
...
+		kind = strings.ToLower(gvk.Kind)
+	}
+	statusUpdatesTotal.WithLabelValues(string(result), kind).Inc()
```
lukas016, line 40: i recommend use https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#CounterVec.GetMetricWith instead of WithLabelValues.

Reason: WithLabelValues is very heavy depending on order of labels. So later changes can affects label logic (order was changed) and it is very hard catch that during review.
Second problem which is more important: Function for bad labels will trigger panic. This panic can affect business logic. Personally prefer handle error myself and just log it and continue in business logic.

api/internal/handler/util/getters.go
lukas016: or you can use sort function with UID which will help you write simpler logic

common/pkg/controller/metric.go
ron96g: But this will then be visible in the metrics anyways because of "unknown"

lukas016: yes but you won't know root cause and you will need to add for debugging this logging anyway.

api/internal/handler/util/getters.go
ron96g: fixed using cmp.Or(...)

common/pkg/controller/metric.go
ron96g: We have a wrapper around this metrics anyways so as long as nobody ignores it, there is no risk: `recordStatusUpdate`

lukas016: yes but you will create bad example for AI.

julius-malcovsky (approved):

PR #544, opened 2026-08-06
fix(common): always requeue reconcilation after FirstSetup

lukas016:
When FirstSetup modifies an object (e.g. adds a finalizer or initializes conditions), the controller was returning reconcile.Result{} — effectively stopping reconciliation. This relied on the generation predicate to trigger a follow-up reconcile, but the generation predicate only fires on spec changes, not on metadata or status changes produced by FirstSetup.

As a result, objects could get stuck after initial setup — the finalizer and conditions would be written, but the full reconciliation loop would never run.

Fix by returning Requeue: true after FirstSetup reports a change, so the object is always reconciled again immediately regardless of which predicates are registered on the controller.


ron96g (approved):
Looks good

PR #545, opened 2026-08-06
fix(common): preserve original error when status update fails

lukas016:
When the status update after a reconcile error also failed, the original error was discarded and replaced by the status update error. This caused the original failure to be silently lost, masking the root cause.

Join both errors so the controller returns and logs all failure context.

### DIFF
--- a/common/pkg/controller/controller.go
+++ b/common/pkg/controller/controller.go
@@ -6,6 +6,7 @@ package controller
 
 import (
 	"context"
+	"errors"
 	"reflect"
 
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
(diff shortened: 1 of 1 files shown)


stefan-ctrl (changes requested):
Hi,
Thanks for the suggestion! The identified and described issue looks valid. Have you tested it somewhere whether it works as expected?
I'm missing unit tests, so it's hard for me to verify right now.

Also, I fear that joining errors as suggested in the commits would end up in stacking errors indefinitely (something that was an issue in the old controlplane and would overload the operators as well as k8s).
Hence, restrictions would need to be set (i.e., only save the first and the latest issue, or more sophisticated alternatives). 
Any thoughts on that?

Based on the missing tests and the expectation that errors would be stacked indefinitely, I do not feel confident in merging it right now.

stefan-ctrl (dismissed):

stefan-ctrl (approved):

PR #546, opened 2026-08-06
feat(sftp): optimize error handling

lukas016:
- use standard error instead of Retryable
- ignore not found error in deletion operations

### DIFF
--- a/sftp/internal/handler/instance/handler.go
+++ b/sftp/internal/handler/instance/handler.go
@@ -74,7 +74,7 @@ func (h *InstanceHandler) Delete(ctx context.Context, obj *sftpv1.Instance) erro
 	}
 
 	err = sftpService.DeleteSFTPUser(ctx, obj.Name)
-	if err != nil {
+	if err != nil && !errors.Is(err, service.ErrNotFound) {
 		return fmt.Errorf("deleting SFTP user %q: %w", obj.Name, err)
 	}
 
--- a/sftp/internal/service/api_error.go
+++ b/sftp/internal/service/api_error.go
@@ -36,6 +36,7 @@ func handleAPIError(operation string, statusCode int, body []byte, apiErr *ApiEr
 	errMessage := fmt.Sprintf("SFTP Tardis API returned %d while trying to %s: %s", statusCode, operation, message)
 	switch {
 	case statusCode == http.StatusBadRequest:
+		// bad request represent error on client side, so we return a blocked error to avoid retrying
 		return ctrlerrors.BlockedErrorf("%s", errMessage)
 	case statusCode == http.StatusNotFound:
 		return fmt.Errorf("%w: %s", ErrNotFound, errMessage)
--- a/sftp/internal/service/api_error.go
+++ b/sftp/internal/service/api_error.go
@@ -5,12 +5,15 @@
 package service
 
 import (
+	"errors"
 	"fmt"
 	"net/http"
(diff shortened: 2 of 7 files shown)


sftp/internal/handler/instance/handler.go
```diff
@@ -74,7 +74,7 @@ func (h *InstanceHandler) Delete(ctx context.Context, obj *sftpv1.Instance) erro
 	}
 
 	err = sftpService.DeleteSFTPUser(ctx, obj.Name)
-	if err != nil {
+	if err != nil && !errors.Is(err, service.ErrNotFound) {
 		return fmt.Errorf("deleting SFTP user %q: %w", obj.Name, err)
 	}
```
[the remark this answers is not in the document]
lukas016: fixed

sftp/internal/service/api_error.go
```diff
@@ -32,15 +35,17 @@ func handleAPIError(operation string, statusCode int, body []byte, apiErr *ApiEr
...
+		statusCode == http.StatusForbidden:
+		return errors.New(errMessage)
 	case statusCode >= http.StatusInternalServerError:
-		return ctrlerrors.RetryableErrorf("%s", errMessage)
+		return errors.New(errMessage)
 	default:
-		return ctrlerrors.RetryableErrorf("%s", errMessage)
+		return errors.New(errMessage)
 	}
 }
```
stefan-ctrl: Could you add some comments on why certain errors are no longer wrapped around `ctrlerrors.BlockedErrorf`?

sftp/internal/service/http_service.go
```diff
@@ -58,7 +56,7 @@ func newClientWithResponses(cfg Config) (ClientWithResponsesInterface, error) {
 func (s *HTTPService) CreateOrUpdateSFTPUser(ctx context.Context, user RoverSftpUserModel) error {
 	res, err := s.client.CreateOrUpdateSftpUserWithResponse(ctx, user)
 	if err != nil {
-		return ctrlerrors.RetryableErrorf("SFTP Tardis API request failed: %s", err.Error())
+		return fmt.Errorf("SFTP Tardis API request failed: %w", err)
```
stefan-ctrl: Similar to the previous comment: Why should this not be a retryable error? 
Adding a comment in this PR would help understand why this optimizes the error handling.

stefan-ctrl (changes requested):
Code wise very good. 
But would like to understand why these changes optimize the error handling.

sftp/internal/service/http_service.go
lukas016: Technically it is still retryable error but main difference is: this is using exponential delay interval for reconciliation. Retryable error are using static interval which is 1s as default and it consume more CPU as it is needed.

Do you really need some comment here? Because it is related with changes in common/pkg/controller and technically it just reflect behavior of common package.

sftp/internal/service/api_error.go
lukas016: fixed && pushed

sftp/internal/service/http_service.go
stefan-ctrl: Talked about iut off-github, and demonstrated the impact of the change. Resolved and accepted :)

stefan-ctrl (approved):
Demonstrated the changes and did a quick chat about it. Approved 👍

PR #555, opened 2026-08-07
feat: reworked zone structure; added consumer-failover tokenUrl

ron96g:


admin/api/Makefile
```diff
@@ -0,0 +1,14 @@
...
+
+.PHONY: build fmt vet test
+
+fmt:
+	go fmt ./...
+
+vet:
+	go vet ./...
+
+test: fmt vet
```
stefan-ctrl: Will this test be executed in the pipelines as part of the `admin` jobs?

docs/docs/admin-journey/features/failover.md
```diff
@@ -41,49 +40,78 @@ metadata:
...
+          enabled: true
   # ... rest of zone configuration
 ```
 
 ### Configuration Details
 
 | Field | Description |
 | ----- | ----------- |
-| `presets[].name` | Must be a valid identifier. The preset that enables consumer failover is typically named `ConsumerFailover`. |
+| `presets[].name` | Must be a valid identifier. The preset that enables consumer failover is typically named `consumer-failover`. |
```
stefan-ctrl: Is the example with failover really the case? We may want to remove that.

admin/api/v1/zone_types.go
```diff
@@ -206,52 +236,27 @@ func (p *GatewayConfigPreset) GetDefaultUrl() string {
...
-	return nil, fmt.Errorf("%w: %s", ErrNoPresetFound, name)
-}
+const (
+	// GatewayTypeAPI serves synchronous API traffic (exposures, subscriptions, proxy routes).
+	GatewayTypeAPI GatewayType = "API"
+	// GatewayTypeAI serves agentic traffic (MCP servers, agent cards).
+	GatewayTypeAI GatewayType = "AI"
+	// GatewayTypeEvent serves asynchronous event traffic.
+	GatewayTypeEvent GatewayType = "Event"
+)
```
stefan-ctrl: Question:
Let's say we merge it: How would that affect the development of SFTP and file domains?

Should we perhaps postpone the merge of this refactor until file/sftp is ready and merged?

stefan-ctrl: I'd propose they do not have to affect each other, and SFTP/FILE should work without checking for zones present and just pick the default for now, and later add a new GatewayType if required, so as to keep the changes in the file/sftp to a minimum.

stefan-ctrl (commented):
LGTM.
Looked over the changes since the talk last Thuesday. I'd be confident merging it, once some questions are resolved regarding _when_ to merge. 

Best we talk about it in one of our dailies.

admin/api/Makefile
ron96g: No, I think this is a general issue that api module tests are not being executed.

docs/docs/admin-journey/features/failover.md
ron96g: Its just an example on how to name it

admin/api/v1/zone_types.go
ron96g: it will affect it. Question is twhat is merged first :D

PR #560, opened 2026-08-10
ci: add mirror_to_artifactory job using skopeo

BjoernKarma:


BjoernKarma:
In the team we decided that only release versions are mirrored. This is then not needed currently. Will close the PR. In case this is needed later, we can re-activate.

PR #561, opened 2026-08-10
feat/install improvements

ron96g:
- **feat(file-manager): added rustfs as default-backend; deleted minio references**
- **feat(install): align config-handling of all components**


stefan-ctrl (changes requested):
Great change! Only have a few minor comments regarding documentation, such as duplicated information in multiple places.


But otherwise seems fine. Might need to wait for customer feedback once they deploy their own dCP for more specific and outside feedback regarding the docs.

stefan-ctrl (approved):

PR #562, opened 2026-08-11
feat(cpapi): add permissionsURL to application

stefan-ctrl:
# Query

Takes about ~2.5 - 3sec current for all applications in clusters without limit

```
query ApplicationPermissionsURL {
  applications {
    edges {
      node {
        name
        clientID
        permissionsURL
      }
    }
  }
}
```

# Sample Response
```
        {
          "node": {
            "name": "provider-jm-demo-application-new",
            "clientID": "devops--demo-team--example",
            "permissionsURL": "https://xxxxxx.de/eni/chevron/v2/permission?application=devops--demo-team--example",
          }
        },
```

### DIFF
--- a/controlplane-api/schema.graphql
+++ b/controlplane-api/schema.graphql
@@ -279,6 +279,8 @@ type EventExposureInfo {
 extend type Application {
   "Owning team (reduced view for cross-tenant safety)"
   ownerTeam: TeamInfo!
+  "PermissionsUrl for permission queries (dynamically built from gateway URL) with ?application=<clientId> appended when querying"
+  permissionsURL: String @goField(forceResolver: true)
 }
 
 extend type Zone {
--- a/controlplane-api/ent.graphql
+++ b/controlplane-api/ent.graphql
@@ -3741,6 +3741,7 @@ type Zone implements Node {
   name: String!
   gatewayURL: String
   issuerURL: String
+  permissionURL: String
   visibility: ZoneVisibility!
   applications: [Application!]
 }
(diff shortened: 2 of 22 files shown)


controlplane-api/schema.graphql
```diff
@@ -279,6 +279,8 @@ type EventExposureInfo {
 extend type Application {
   "Owning team (reduced view for cross-tenant safety)"
   ownerTeam: TeamInfo!
+  "PermissionsUrl for permission queries (dynamically built from gateway URL) with ?application=<clientId> appended when querying"
+  permissionsURL: String @goField(forceResolver: true)
```
ron96g, line 283: Any reason why we need to extend the schema here? Could this done via the Application schema and then the projector adds this? Yes, it would mean more data stored in the DB (but not that much as its a niche feature) but we remove the extra edge traversal from Application to Zone. Especially if we assume that a client (MC) would just use 1 GraphQL-query this would mean a lot of unnecessary traversals to Zone just to return nil. At the very least we should at a conditonal check in the resolver to verify first if the Application actually has Permissions enabled.

ron96g (changes requested):

controlplane-api/schema.graphql
stefan-ctrl: Good point. I implemented it this way to not store duplicated data.
As permission is implemented right now, `permissionURL `would then be stored for every application of the respective zone once the permission feature is enabled at the zone level.
This would result in duplicating the `permissionURL` for every application in that zone, even if only used by a few customers.

But since strings are inexpensive to store,  I will make a second implementation and check the runtime difference of GraphQL.

ron96g: No, the projector needs to check if the Permissions feature is used and only if its used, it would store the URL otherwise it would be null

PR #563, opened 2026-08-11
feat(sftp): support events for instance

lukas016:
use always all events for instance

### DIFF
--- a/sftp/internal/handler/instance/handler.go
+++ b/sftp/internal/handler/instance/handler.go
@@ -20,6 +20,13 @@ import (
 	"github.com/telekom/controlplane/sftp/internal/service"
 )
 
+var events = []service.RoverSftpUserModelHorizonNotificationEvents{
+	service.Delete,
+	service.Download,
+	service.Upload,
+	service.Rename,
+}
+
 var _ handler.Handler[*sftpv1.Instance] = &InstanceHandler{}
 
 type InstanceHandler struct {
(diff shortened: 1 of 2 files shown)


sftp/internal/handler/instance/handler.go
```diff
@@ -82,11 +89,8 @@ func (h *InstanceHandler) Delete(ctx context.Context, obj *sftpv1.Instance) erro
...
 
 func (h *InstanceHandler) createOrUpdateServiceUser(ctx context.Context, sftpService service.Service, instance *sftpv1.Instance) error {
-	events := []service.RoverSftpUserModelHorizonNotificationEvents{}
-
 	sftpUser := service.RoverSftpUserModel{
-		SftpUserName: instance.Name,
-		// nil creates invalid user in an external service
+		SftpUserName:              instance.Name,
 		HorizonNotificationEvents: &events,
 	}
```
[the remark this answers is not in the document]
lukas016: events are marshall into []byte in code. later changes should be covered during review.

sftp/internal/handler/instance/handler.go
```diff
@@ -20,6 +20,13 @@ import (
 	"github.com/telekom/controlplane/sftp/internal/service"
 )
 
+var events = []service.RoverSftpUserModelHorizonNotificationEvents{
+	service.Delete,
+	service.Download,
+	service.Upload,
+	service.Rename,
+}
```
stefan-ctrl: What's the benefit of it being file-global?

stefan-ctrl (commented):
Tbh, I'm not quite sure,  why these changes are needed, and whats the impact of it in the sftp-domain. Mind elaborating on it a bit more?

sftp/internal/handler/instance/handler.go
lukas016: it is use just for reading and variable doesn't have to be initialized for every function again. But golang doesn't support constant for slices, so it has to be variable. Benefits are better performance (reinitialization isn't needed) and it shouldn't affect GC (again performance improvement).

Events are later in function marshall into []body, so it is safe use global variable here. Of course if someone change logic later and it will modify events, it shouldn't be global variable anymore.

stefan-ctrl (approved):
Talked about it off-github and explained the reason for these changes in more depth. Approved.

PR #567, opened 2026-08-12
feat(sftp): expose metrics with service monitor

lukas016:


stefan-ctrl (approved):
LGTM. Compared config with other controllers. 
If you've checked and metrics collection works, can be merged 👍

PR #591, opened 2026-08-18
ci: dependabot npm updates grouped by update type

BjoernKarma:
npm updates are only relevant for the docs directory and docusaurs requires updating dependencies together. This change configures dependabot to group updates by update type combining all minor changes and all major changes together in one PR.


stefan-ctrl (approved):
LGTM. The less we have to deal with JS dependencies, the better 👍

PR #592, opened 2026-08-18
feat(make): enhance tidy-all to include go.mod in api subdirectory

stefan-ctrl:


stefan-ctrl:
Not needed. See https://github.com/telekom/controlplane/actions/runs/32109669927/job/95626179923#step:4:16

PR #602, opened 2026-08-18
ci: harden dependabot-tidy workflow to allow auto-triggering CI

BjoernKarma:
### **pull_request** → **pull_request_target**

The old event fired in the PR's fork/bot context with a read-only token, requiring maintainer approval. **pull_request_target** always runs in the base branch context with a write-capable **GITHUB_TOKEN** no maintainer authorization needed.

Security safeguard added: we check out the PR by its immutable SHA (**pr.head.sha**), not the branch name. This prevents a malicious PR from being substituted mid-run. The job is also guarded by `github.actor == 'dependabot[bot]'`.

### GitHub App token for re-triggering CI

A commit pushed with **GITHUB_TOKEN** is explicitly blocked from re-triggering workflows by GitHub (anti-recursion). A GitHub App token bypasses this restriction. The new step conditionally generates one using `actions/create-github-app-token` when **APP_ID** and **APP_PRIVATE_KEY** secrets exist.  Falls back gracefully to **GITHUB_TOKEN** (push works, but CI won't auto-rerun).

### Explicit push target

Checking out by SHA puts git in detached HEAD state. The old **git push** would fail. Now it uses `git push origin HEAD:"$HEAD_REF"` with the branch name surfaced from the PR metadata.

**workflow_dispatch** now accepts **pr_number** So a maintainer can manually run tidy on any PR without needing to edit the workflow

### DIFF
--- a/.github/workflows/dependabot-tidy.yaml
+++ b/.github/workflows/dependabot-tidy.yaml
@@ -4,12 +4,28 @@
 
 name: Dependabot Tidy
 
+# pull_request_target runs in the context of the BASE branch, which gives the
+# workflow a privileged GITHUB_TOKEN (contents: write) even for Dependabot PRs.
+# We guard the job strictly to dependabot[bot] and always check out the PR
+# head by its immutable SHA (not a mutable branch ref) to prevent privilege
+# escalation from arbitrary PRs.
+#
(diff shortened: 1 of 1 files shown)


stefan-ctrl (approved):
LGTM. Has harding to prevent go mod tidy jobs overwrite other stuff and generally prevent escalation.

Should try it out, if this solves some issues. 👍

PR #610, opened 2026-08-19
feat(cpapi): add rate limiting to ApiSubscriptions

stefan-ctrl:
# Overview

3d24df6b03d45ccdb3a06b97ea878a271335ca76: feat(projector): add rate limiting support to ApiSubscription with traffic configuration

c35e06e756b6baf4e9d8cbe84aa6c4a7f5ab8a40: feat(cpapi): add traffic field to ApiSubscription and related mutations

# Request

```graphql
query ListApiSubscriptionRateLimits {
  apiSubscriptions(first: 50) {
    edges {
      node {
        id
        basePath
        owner { name }
        traffic {
          subscriberLimits { second minute hour }
          providerLimits   { second minute hour }
        }
      }
    }
  }
}
```

# Response with Sample data
```json
{
  "data": {
    "apiSubscriptions": {
      "edges": [
        {
          "node": {
            "id": "1",
            "basePath": "xxx",
            "owner": {
              "name": "xxx-consumer"
            },
            "traffic": null
          }
        },
        {
          "node": {
            "id": "2",
            "basePath": "yyy",
            "owner": {
              "name": "yyy-consumer"
            },
            "traffic": {
              "subscriberLimits": {
                "second": 5,
                "minute": 10,
                "hour": 20
              },
              "providerLimits": {
                "second": 50,
                "minute": 100,
                "hour": 200
              }
            }
          }
        },
        ...
}
```


BjoernKarma (commented):
The PR looks good for adding rate limiting to ApiSubscriptions as we discussed in the team. 

There are only two question that came to my mind:
- What happens if the rate limit is changed on the ApiExposure? I think currently it might be possible that the ApiSubscription is not updated and we end up having a stale information in the ApiSubscription. 
- The same might be happening for other lifecycle events for ApiExposure like "delete ApiExposure" or removing the rate limit completely from the exposure. Could you please check?

julius-malcovsky (approved):
LGTM

BjoernKarma (approved):
LGTM 

(Maybe we can investigate the delete use case later as well? Delete ApiExposure --> Delete Traffic. It is probably also okay, because once the exposure is re-created, the subscription gets updated and until then the subscription is stale anyway)

PR #611, opened 2026-08-19
feat(cpapi): add rate limiting to ApiSubscriptions on graphql query time

stefan-ctrl:
Alternative to #610:
- evaluated on read only, without storing data via projector. Benefit: The projector does not have to check the exposure of every api subscritpion insert


stefan-ctrl:
Team decision on 24.08.2026: Closed in favor of #610. We will start with the "store in projector" approach and see the results from there.

PR #612, opened 2026-08-19
ci: disable config verification in golangci-lint to prevent timeouts

stefan-ctrl:
The golangci-lint CI step sometimes fails and does not continue with linting. 

```
Error: Command failed: golangci-lint config verify
... failing to load "https://golangci-lint.run/jsonschema/golangci.v2.11.jsonschema.json":
context deadline exceeded (Client.Timeout exceeded while awaiting headers)
```

This removes the flaky network dependency from the lint step. Actual linting is unaffected; only the schema-validation of .golangci.yml is skipped.

### DIFF
--- a/.github/workflows/reusable-go-ci.yaml
+++ b/.github/workflows/reusable-go-ci.yaml
@@ -157,6 +157,9 @@ jobs:
         with:
           version: v2.11
           working-directory: ${{ inputs.module }}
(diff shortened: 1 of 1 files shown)


BjoernKarma (approved):
LGTM - maybe in the future we can have a look at one-time schema validation of the golang-ci-lint config changes. But it works currently and I think it is okay to not run a validation on every execution.

PR #619, opened 2026-08-21
feat(cpapi): Added owner Application on Api and EvenType entities

iagarba:


stefan-ctrl (changes requested):

controlplane-api/schema.graphql
```diff
@@ -19,6 +19,16 @@ type TeamInfo {
   description: String
 }
 
+type OwnerApplicationInfo {
+  applicationId: ID!
+  externalIds: [OwnerApplicationExternalId!]!
+}
```
stefan-ctrl: Team review note: Should be `id` instead of `applicationId`. 

```go
type OwnerApplicationInfo {
  applicationId: ID!
  externalIds: [OwnerApplicationExternalId!]!
}
```

PR #620, opened 2026-08-21
…docs: Update workflows document with recent additions

BjoernKarma:
docs: Update workflows document with recent additions for roverctl image, image mirroring and openssf scorecard


stefan-ctrl (commented):
Great addition to the docs, so its not getting outdated.

Have two minor comments, but I'm not sure how to proceed there. Best we talk about that in our daily stand-up.

iagarba (approved):

PR #625, opened 2026-08-25
feat(agentic-cpapi): agnetic CRs are projected and prepared for qgl

julius-malcovsky:


controlplane-api/ent/schema/application.go
```diff
@@ -95,6 +95,10 @@ func (Application) Edges() []ent.Edge {
 			Annotations(entgql.RelayConnection()),
 		edge.To("subscribed_events", EventSubscription.Type).
 			Annotations(entgql.RelayConnection()),
+		edge.To("agentic_exposures", AgenticExposure.Type).
+			Annotations(entgql.Skip(entgql.SkipType)),
+		edge.To("agentic_subscriptions", AgenticSubscription.Type).
+			Annotations(entgql.Skip(entgql.SkipType)),
```
stefan-ctrl: Minor:
From a naming convention point of view: `exposed_agentics` and `subscribed_agentics` would fit with the others.

projector/internal/domain/agenticsubscription/translator.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	var traffic *model.AgenticSubscriberTraffic
+	if obj.Spec.Traffic.Failover != nil {
+		traffic = &model.AgenticSubscriberTraffic{
+			Failover: &model.AgenticSubscriberFailover{
+				Enabled: obj.Spec.Traffic.Failover.Enabled,
+			},
+		}
+	}
```
stefan-ctrl: Question:
Is it by design that AgenticSubs can not view their rate limits (i.e., `*model.AgenticSubscriberTraffic` does not contain any rate-limiting information)?

ApiSubscription has this information. Just wanted to confirm whether this needs to be added or is not required.

stefan-ctrl (commented):
Generally, looks solid.
Two minor things, once resolved, I'd be happy to approve.

projector/internal/domain/agenticsubscription/translator.go
julius-malcovsky: Good question. I think we decided that agentic will be without features, even though not all are yet removed/blocked. Problem is the large duplication of codebase between api and agentic and we didnt find a suitable solution for that yet

controlplane-api/ent/schema/application.go
julius-malcovsky: good call, thanks updated

projector/internal/domain/agenticsubscription/translator.go
julius-malcovsky: also, in this case, im not 100% if for example rate limit even makes sense .. since agentic items mostly work as streams and rate limit counts individual requests

stefan-ctrl (approved):

